PR #1, opened 2017-10-26
OCI-Monitor integration fixes

zoxpx:
* added WatchNode() that can install watcher functions to Node's changes
* extended SearchNodeByAddresses() with 2 more failbacks when looking for Nodes
* removed dependency to torpedo/pkg/task package
* initK8sClient() now checks the connectivity to Kubernetes server

### DIFF
--- a/k8s/k8s.go
+++ b/k8s/k8s.go
@@ -7,7 +7,6 @@ import (
 	"sync"
 	"time"
 
-	"github.com/portworx/torpedo/pkg/task"
 	meta_v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/kubernetes/typed/apps/v1beta1"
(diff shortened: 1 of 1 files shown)


k8s/k8s.go
```diff
@@ -349,6 +397,53 @@ func (k *k8sOps) RemoveLabelOnNode(name, key string) error {
...
+		return err
+	}
+
+	// fire off watch function
+	go func() {
+		for {
+			select {
+			case event, more := <-watchInterface.ResultChan():
+				if !more {
+					// log.Warn("Kubernetes node watch channel closed")
```
harsh-px: Closing of the watch channel needs to be handled since this has been seen on QA setups very frequently. Retries were added in porx to perform this watch. Referring to below function in porx k8s_node_utils.go

You can check this version in and add the retries in a subsequent commit with a TODO here.

```
func (knu *k8sNodeUtils) handleWatch(
	watchInterface k8s_watch.Interface,
	watchNodeFn NodeWatchFunc,
	nodeHostname string,
) {
	for {
		select {
		case event, more := <-watchInterface.ResultChan():
			if !more {
				var err error
				for retries := 0; retries < maxKubernetesWatchRetries; retries++ {
					// We lost the watch connection most probably because kube api-server is down.
					// Sleep for sometime and retry again.
					time.Sleep(5 * time.Second)
					err = knu.WatchNode(nodeHostname, watchNodeFn)
					if err != nil {
						continue
					}
					break
				}
				if err != nil {
					logrus.Warn("Kubernetes node watch "+
						"channel closed. Failed to restart watch. Error: %v", err)

				}
				return
			}
			k8sNode, ok := event.Object.(*k8s_types.Node)
			if ok {
				watchNodeFn(k8sNode)
			}
		}
	}
}
```

zoxpx: Why can't we check in this version "as is" -- and have Aditya fix up the function properly as he is the original author.  I just ported the function from API v1.5 to v4.0.

harsh-px: Sure, go ahead.

PR #3, opened 2017-10-30
Add package for task retry function

harsh-px:


task/task.go
```diff
@@ -0,0 +1,53 @@
+package task
+
+import (
+	"errors"
+	"time"
+
+	"github.com/sirupsen/logrus"
```
zoxpx: Could you please correct this to `github.com/Sirupsen/logrus`  (capital 'S')? -- don't bother w/ new PR
This is currently breaking the PX-Installer, and besides, the logrus author recommends to use the package with the "capital 'S'".

PR #6, opened 2017-11-13
Using objects from k8s.io/api instead of k8s.io/client-go

piyush-nimbalkar:
k8s.io/api is going to be the canonical place for all api objects going forward

### DIFF
--- a/k8s/k8s.go
+++ b/k8s/k8s.go
@@ -8,12 +8,12 @@ import (
 	"time"
 
 	"github.com/portworx/sched-ops/task"
+	apps_api "k8s.io/api/apps/v1beta1"
+	"k8s.io/api/core/v1"
+	storage_api "k8s.io/api/storage/v1"
 	meta_v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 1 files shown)


k8s/k8s.go
```diff
@@ -252,8 +252,7 @@ func (k *k8sOps) IsNodeReady(name string) error {
 		case v1.NodeConditionType(v1.NodeOutOfDisk),
 			v1.NodeConditionType(v1.NodeMemoryPressure),
 			v1.NodeConditionType(v1.NodeDiskPressure),
-			v1.NodeConditionType(v1.NodeNetworkUnavailable),
-			v1.NodeConditionType(v1.NodeInodePressure):
```
piyush-nimbalkar: had to remove this, as it is not present in the 1.8.2 version of v1 api

disrani-px:
Can we also create issues to vendor the dependencies and add a Makefile to prevent build breaks.

piyush-nimbalkar:
https://github.com/portworx/sched-ops/issues/7
Create issue to track addition of Makefile and vendor directory

PR #11, opened 2017-12-09
Add snapshot interface

disrani-px:


k8s/k8s.go
```diff
@@ -1137,50 +1160,144 @@ func (k *k8sOps) isPVCShared(pvc *v1.PersistentVolumeClaim) bool {
...
+		Name(snap.Metadata.Name).
+		Resource(snap_v1.VolumeSnapshotResourcePlural).
+		Namespace(snap.Metadata.Namespace).
+		Do().Error()
+}
+
+func (k *k8sOps) ValidateSnapshot(snap *snap_v1.VolumeSnapshot) error {
+	t := func() (interface{}, error) {
+		var result snap_v1.VolumeSnapshot
+		if err := k.snapClient.Get().
```
adityadani: k.GetSnapshotStatus()

k8s/k8s.go
```diff
@@ -1137,50 +1160,144 @@ func (k *k8sOps) isPVCShared(pvc *v1.PersistentVolumeClaim) bool {
...
+			Name(snap.Metadata.Name).
+			Resource(snap_v1.VolumeSnapshotResourcePlural).
+			Namespace(snap.Metadata.Namespace).
+			Do().Into(&result); err != nil {
+			return "", err
+		}
+
+		for _, condition := range result.Status.Conditions {
+			if condition.Type == snap_v1.VolumeSnapshotConditionReady && condition.Status == v1.ConditionTrue {
```
adityadani: Instead of trying it for 5 minutes shouldn't we just stop checking if we get VolumeSnapshotConditionError as true

k8s/k8s.go
```diff
@@ -1137,50 +1160,144 @@ func (k *k8sOps) isPVCShared(pvc *v1.PersistentVolumeClaim) bool {
...
+			}
+		}
+
+		return "", &ErrSnapshotNotReady{
+			ID:    result.Metadata.Name,
+			Cause: fmt.Sprintf("Snapshot Status %v", result.Status),
+		}
+	}
+
+	if _, err := task.DoRetryWithTimeout(t, 5*time.Minute, 10*time.Second); err != nil {
```
adityadani: nit: define as constants.

k8s/k8s.go
disrani-px: Done

k8s/k8s.go
disrani-px: I think Piyush was going to make this configurable for all the timeouts. Leaving as is for now.

k8s/k8s.go
disrani-px: Update the retry task to stop if the task function returns false for retry

k8s/k8s.go
```diff
@@ -1137,50 +1160,162 @@ func (k *k8sOps) isPVCShared(pvc *v1.PersistentVolumeClaim) bool {
...
+		Resource(snap_v1.VolumeSnapshotResourcePlural).
+		Namespace(snap.Metadata.Namespace).
+		Do().Error()
+}
+
+func (k *k8sOps) ValidateSnapshot(snap *snap_v1.VolumeSnapshot) error {
+	if err := k.initK8sClient(); err != nil {
+		return err
+	}
+	t := func() (interface{}, error) {
```
adityadani: Shouldnt this be -> t := func() (interface{}, bool, error)

adityadani: Also at all other places where DoWithRetryTimeout is used.

k8s/errors.go
```diff
@@ -67,3 +67,27 @@ type ErrPVCNotReady struct {
...
+	Cause string
+}
+
+func (e *ErrSnapshotNotReady) Error() string {
+	return fmt.Sprintf("Snapshot %v is not ready yet. Cause: %v", e.ID, e.Cause)
+}
+
+// ErrSnapshotFailed error type for when a snapshot has failed
+type ErrSnapshotFailed struct {
+	// ID is the identifier of the app
```
piyush-nimbalkar: same as above?

k8s/k8s.go
```diff
@@ -158,21 +161,35 @@ type PersistentVolumeClaimOps interface {
...
 	// GetPersistentVolumeClaimParams fetches custom parameters for the given PVC
 	GetPersistentVolumeClaimParams(*v1.PersistentVolumeClaim) (map[string]string, error)
 	// GetPersistentVolumeClaimStatus returns the status of the given pvc
 	GetPersistentVolumeClaimStatus(*v1.PersistentVolumeClaim) (*v1.PersistentVolumeClaimStatus, error)
 }
 
+type SnapshotOps interface {
+	// CreateSnapshot creates the given snapshot
+	CreateSnapshot(*snap_v1.VolumeSnapshot) (*snap_v1.VolumeSnapshot, error)
+	// DeletePersistentVolumeClaim deletes the given snapshot
```
piyush-nimbalkar: *DeleteSnapshot

k8s/k8s.go
```diff
@@ -158,21 +161,35 @@ type PersistentVolumeClaimOps interface {
...
+	DeleteSnapshot(*snap_v1.VolumeSnapshot) error
```
piyush-nimbalkar: I think as an interface we should be passing the name and namespace in delete calls. We did in some places but it is pending in other places. 
The advantage is when we don't have the object, but given just the name and namespace you can still delete it. We don't have to get the object and then pass it to the delete. What do you think?

k8s/k8s.go
```diff
@@ -1137,50 +1160,162 @@ func (k *k8sOps) isPVCShared(pvc *v1.PersistentVolumeClaim) bool {
...
+
+	var result snap_v1.VolumeSnapshot
+	if err := k.snapClient.Get().
+		Name(snap.Metadata.Name).
+		Resource(snap_v1.VolumeSnapshotResourcePlural).
+		Namespace(snap.Metadata.Namespace).
+		Do().Into(&result); err != nil {
+		return "", err
+	}
+	return result.Metadata.Name, nil
```
piyush-nimbalkar: Is the snapshot name same as volume name? If yes, why do we call the method getVolumeForSnapshot instead of getSnapshotName or directly use the snapshot.metadata.name wherever we are using it?

task/task.go
```diff
@@ -13,11 +13,12 @@ import (
 var ErrTimedOut = errors.New("timed out performing task")
 
 // DoRetryWithTimeout performs given task with given timeout and timeBeforeRetry
-func DoRetryWithTimeout(t func() (interface{}, error), timeout, timeBeforeRetry time.Duration) (interface{}, error) {
+func DoRetryWithTimeout(t func() (interface{}, bool, error), timeout, timeBeforeRetry time.Duration) (interface{}, error) {
```
piyush-nimbalkar: I remember Zoran was using this method somewhere other than stork and torpedo; and probably a couple of places in torpedo as well.
My recommendation is, if possible have a separate method wrapping the original one with the retry boolean. That way we don't have change the original consumers when they upgrade to the latest sched-ops. Else we just keep a different method altogether to avoid changing this in multiple projects.

k8s/k8s.go
```diff
@@ -186,7 +203,7 @@ func Instance() Ops {
 // Initialize the k8s client if uninitialized
 func (k *k8sOps) initK8sClient() error {
 	if k.client == nil {
-		k8sClient, err := getK8sClient()
+		k8sClient, snapClient, err := getK8sClient()
```
piyush-nimbalkar: why not have a separate method to get k8sSnapClient? currently it seems both are not used at the same time. We can call initSnapClient for snapshot methods. That way failures in initializing snapclient won't affect the k8s client.

k8s/k8s.go
disrani-px: Done

k8s/errors.go
```diff
@@ -67,3 +67,27 @@ type ErrPVCNotReady struct {
 func (e *ErrPVCNotReady) Error() string {
 	return fmt.Sprintf("PVC %v is not ready yet. Cause: %v", e.ID, e.Cause)
 }
+
+// ErrSnapshotNotReady error type for when a snapshot is not yet ready/bound
+type ErrSnapshotNotReady struct {
+	// ID is the identifier of the app
```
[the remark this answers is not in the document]
disrani-px: Done

k8s/errors.go
disrani-px: Done

k8s/k8s.go
disrani-px: Yes, definitely better to just use name and namespace. I copied the interface from volumes for these. I'll update it

k8s/k8s.go
disrani-px: I think it's better to have just one function to initialize everything. Eventually the snapClient will go away once the snap changes in k8s are made upstream. So we will have to just change initK8sClient() instead of changing the init calls from Snapshot functions..

k8s/k8s.go
disrani-px: It doesn't have to be the same. Right now stork uses the snapshot name, which could change. This creates an abstraction layer.

task/task.go
piyush-nimbalkar: Up to you. Just communicate to Zoran, and probably change in torpedo because torpedo used sched-ops master as of now.

PR #12, opened 2017-12-14
Portworx 1.3.0 port

zoxpx:
* Adding FindMyNode() and GetLocalIPList() utility methods from oci-monitor
* Adding SearchPodsByUID() from porx
* Adding Godeps.json to track package dependencies

### DIFF
--- a/k8s/k8s.go
+++ b/k8s/k8s.go
@@ -34,6 +34,11 @@ const (
 	labelUpdateMaxRetries = 5
 )
 
+var (
+	// ErrPodsNotFound error returned when pod or pods could not be found
+	ErrPodsNotFound = fmt.Errorf("Pod(s) not found")
+)
+
 // Ops is an interface to perform any kubernetes related operations
 type Ops interface {
 	NamespaceOps
@@ -989,7 +994,7 @@ func (k *k8sOps) GetPodByUID(uid string, namespace string) (*v1.Pod, error) {
 		}
 	}
 
-	return nil, nil
+	return nil, ErrPodsNotFound
 }
 
 func (k *k8sOps) IsPodRunning(pod v1.Pod) bool {
--- a/k8s/k8s.go
+++ b/k8s/k8s.go
@@ -2,8 +2,10 @@ package k8s
 
 import (
 	"fmt"
+	"net"
 	"os"
 	"regexp"
+	"strings"
 	"sync"
 	"time"
 
@@ -17,6 +19,7 @@ import (
 	meta_v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/selection"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/watch"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/kubernetes/typed/apps/v1beta2"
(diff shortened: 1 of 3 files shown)


k8s/k8s.go
```diff
@@ -135,6 +140,8 @@ type PodOps interface {
 	GetPods(string) (*v1.PodList, error)
 	// GetPodsByOwner returns pods for the given owner and namespace
 	GetPodsByOwner(string, string) ([]v1.Pod, error)
+	// SearchPodsByUID returns pod with the given UID, or NIL if nothing found
+	SearchPodsByUID(string, string) (*v1.Pod, error)
```
piyush-nimbalkar: nit: SearchPodByUID or rather it could be named GetPodByUID

k8s/k8s.go
```diff
@@ -1321,3 +1353,40 @@ func loadClientFor(config *rest.Config) (*kubernetes.Clientset, *rest.RESTClient
...
+func GetLocalIPList(includeHostname bool) ([]string, error) {
+	ifaces, err := net.Interfaces()
+	if err != nil {
+		return nil, err
+	}
+	ipList := make([]string, 0, len(ifaces))
+	for _, i := range ifaces {
+		addrs, err := i.Addrs()
+		if err != nil {
+			return ipList, fmt.Errorf("Error listing addresses for %s: %s", i.Name, err)
```
piyush-nimbalkar: We should either return nil in the ipList here or If we want to send IP addresses of possible interfaces then we should have a continue instead of return here.

k8s/k8s.go
```diff
@@ -1321,3 +1353,40 @@ func loadClientFor(config *rest.Config) (*kubernetes.Clientset, *rest.RESTClient
 func roundUpSize(volumeSizeBytes int64, allocationUnitBytes int64) int64 {
 	return (volumeSizeBytes + allocationUnitBytes - 1) / allocationUnitBytes
 }
+
+// GetLocalIPList returns the list of local IP addresses, and optionally includes local hostname.
+func GetLocalIPList(includeHostname bool) ([]string, error) {
```
piyush-nimbalkar: Can we make this private if we are not using it externally?

k8s/k8s.go
```diff
@@ -960,6 +976,22 @@ func (k *k8sOps) GetPodsByOwner(ownerName string, namespace string) ([]v1.Pod, e
...
+	}
+
+	pUID := types.UID(uid)
+	for _, pod := range pods.Items {
+		if pod.UID == pUID {
+			return &pod, nil
+		}
+	}
+
+	return nil, nil
```
piyush-nimbalkar: Should we return an error if the pod is not found?

k8s/k8s.go
```diff
@@ -360,6 +367,15 @@ func (k *k8sOps) SearchNodeByAddresses(addresses []string) (*v1.Node, error) {
 	return nil, fmt.Errorf("failed to find k8s node for given addresses: %v", addresses)
 }
 
+// FindMyNode finds LOCAL Node in Kubernetes cluster.
+func (k *k8sOps) FindMyNode() (*v1.Node, error) {
+	ipList, err := GetLocalIPList(true)
+	if err != nil {
```
piyush-nimbalkar: should we also check if the ipList is not empty here?

k8s/k8s.go
zoxpx: We could -- but in that case we better be consistent, and make all other Get/Search -functions (ie. `GetPodsByOwner`) also return the `ErrPodsNotFound` error when stuff isn't found;  note that this also means people will need to check against public Error-object (ie. `if err == ErrPodsNotFound`) instead of empty return values.

k8s/k8s.go
zoxpx: Don't think it's necessary -- 99.999% of the cases this will be a void check, and `k.SearchNodeByAddresses()` already deals with empty lists.

zoxpx:
@piyush-nimbalkar @harsh-px -- since both of you were insisting to have `GetPodByUID()` return `ErrPodsNotFound` if no pod found, I added 47bbaee.
For the sake of consistency, please check other `Get/List*Pod()` methods, and make them return the `ErrPodsNotFound` instead of empty objects/lists.

piyush-nimbalkar:
LGTM
I was not insisting on returning an error, but just wanted to get an opinion on whether we should follow that practice of returning errors on empty Gets as we do for OSD Inspect.

PR #18, opened 2018-01-24
Delete pods in foreground, don't override pod termination grace period and add wait for pod deletion API

harsh-px:


harsh-px:
@zoxpx done. Thanks for reviewing.

PR #19, opened 2018-01-24
Add create and update secret calls

piyush-nimbalkar:


k8s/k8s.go
```diff
@@ -214,6 +214,12 @@ type SnapshotOps interface {
 type SecretOps interface {
 	// GetSecret gets the secrets object given its name and namespace
 	GetSecret(name string, namespace string) (*v1.Secret, error)
+	// CreateSecret creates the given secret
+	CreateSecret(*v1.Secret) (*v1.Secret, error)
+	// UpdateSecret updates the gives secret
+	UpdateSecret(*v1.Secret) (*v1.Secret, error)
+	// UpdateSecretData updates or creates a new secret with the given data
+	UpdateSecretData(string, string, map[string][]byte) (*v1.Secret, error)
```
adityadani: This API is a bit confusing. Let the callers decide what they want to do with the data instead of you just adding key value pairs. You have already provided APIs to get and update secrets.

piyush-nimbalkar: did not want the caller to do the merging of keys-logic; 
also this handles  - create if the secret does not exist else update
Could have called it UpdateSecret, but go doesn't support function overloading.

adityadani: Ok may be call it as AddSecretData then?
Also ideally an Update operation should fail if there does not exist an object.

piyush-nimbalkar: Sure I will call it AddSercretData so that there is no confusion.
Called it update just following the kvdb semantic of Put()

PR #31, opened 2018-02-09
add daemonset operations

harsh-px:

### DIFF
--- a/k8s/k8s.go
+++ b/k8s/k8s.go
@@ -160,14 +160,14 @@ type DaemonSetOps interface {
 	// CreateDaemonSet creates the given daemonset
 	CreateDaemonSet(ds *apps_api.DaemonSet) (*apps_api.DaemonSet, error)
 	// ListDaemonSets lists all daemonsets in given namespace
-	ListDaemonSets(namespace string, listOpts meta_v1.ListOptions) (*apps_api.DaemonSetList, error)
+	ListDaemonSets(namespace string, listOpts meta_v1.ListOptions) ([]apps_api.DaemonSet, error)
 	// GetDaemonSet gets the the daemon set with given name
 	GetDaemonSet(string, string) (*apps_api.DaemonSet, error)
 	// ValidateDaemonSet checks if the given daemonset is ready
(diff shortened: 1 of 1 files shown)


k8s/k8s.go
```diff
@@ -157,14 +157,20 @@ type DeploymentOps interface {
...
+	ListDaemonSets(namespace string, listOpts meta_v1.ListOptions) (*apps_api.DaemonSetList, error)
 	// GetDaemonSet gets the the daemon set with given name
 	GetDaemonSet(string, string) (*apps_api.DaemonSet, error)
 	// ValidateDaemonSet checks if the given daemonset is ready
 	ValidateDaemonSet(string, string) error
 	// GetDaemonSetPods returns list of pods for the daemonset
 	GetDaemonSetPods(*apps_api.DaemonSet) ([]v1.Pod, error)
-	// UpdateDaemonSet updates the given daemon set
-	UpdateDaemonSet(*apps_api.DaemonSet) error
+	// UpdateDaemonSet updates the given daemon set and returns the update ds
```
piyush-nimbalkar: nit: *updated

k8s/k8s.go
```diff
@@ -1008,6 +1014,22 @@ func (k *k8sOps) GetDeploymentsUsingStorageClass(scName string) ([]apps_api.Depl
...
 
+func (k *k8sOps) CreateDaemonSet(ds *apps_api.DaemonSet) (*apps_api.DaemonSet, error) {
+	if err := k.initK8sClient(); err != nil {
+		return nil, err
+	}
+
+	return k.client.Apps().DaemonSets(ds.Namespace).Create(ds)
+}
+
+func (k *k8sOps) ListDaemonSets(namespace string, listOpts meta_v1.ListOptions) (*apps_api.DaemonSetList, error) {
```
piyush-nimbalkar: nit: in most of sched ops calls we return a slice of Objects instead of the k8s List object. should we send the same here so callers can just iterate on it instead of worrying about the metadata?

harsh-px: Yes, returning list is better. Done in incremental.

PR #40, opened 2018-03-25
add helpers for pv and pvc to pods mapping

harsh-px:
This will be used currently in torpedo to figure out pods using a volume.

### DIFF
--- a/k8s/k8s.go
+++ b/k8s/k8s.go
@@ -213,6 +213,10 @@ type PodOps interface {
 	GetPods(string) (*v1.PodList, error)
 	// GetPodsByOwner returns pods for the given owner and namespace
 	GetPodsByOwner(types.UID, string) ([]v1.Pod, error)
+	// GetPodsUsingPV returns all pods in cluster using given pv
+	GetPodsUsingPV(pvName string) ([]v1.Pod, error)
+	// GetPodsUsingPVC returns all pods in cluster using given pvc
+	GetPodsUsingPVC(pvcName, pvcNamespace string) ([]v1.Pod, error)
 	// GetPodsUsingVolumePlugin returns all pods who use PVCs provided by the given volume plugin
 	GetPodsUsingVolumePlugin(plugin string) ([]v1.Pod, error)
 	// GetPodsUsingVolumePluginByNodeName returns all pods who use PVCs provided by the given volume plugin on the given node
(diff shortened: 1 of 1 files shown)


k8s/k8s.go
```diff
@@ -255,6 +259,12 @@ type PersistentVolumeClaimOps interface {
 	ValidatePersistentVolumeClaim(*v1.PersistentVolumeClaim) error
 	// GetPersistentVolumeClaim returns the PVC for given name and namespace
 	GetPersistentVolumeClaim(pvcName string, namespace string) (*v1.PersistentVolumeClaim, error)
+	// GetPersistentVolumeClaims returns all PVCs in given namespace
+	GetPersistentVolumeClaims(namespace string) (*v1.PersistentVolumeClaimList, error)
+	// GetPersistentVolume returns the PV for given name and namespace
```
piyush-nimbalkar: nit: for given name (only)

k8s/k8s.go
```diff
@@ -1788,6 +1829,30 @@ func (k *k8sOps) GetPersistentVolumeClaim(pvcName string, namespace string) (*v1
...
 
+func (k *k8sOps) GetPersistentVolumeClaims(namespace string) (*v1.PersistentVolumeClaimList, error) {
+	if err := k.initK8sClient(); err != nil {
+		return nil, err
+	}
+
+	return k.client.Core().PersistentVolumeClaims(namespace).List(meta_v1.ListOptions{})
+}
+
+func (k *k8sOps) GetPersistentVolume(pvName string) (*v1.PersistentVolume, error) {
```
piyush-nimbalkar: question: aren't PVs also namespaced? if yes, why don't the k8s api need the namespace to get or list the PVs?

k8s/k8s.go
```diff
@@ -1788,6 +1829,30 @@ func (k *k8sOps) GetPersistentVolumeClaim(pvcName string, namespace string) (*v1
...
+	if err := k.initK8sClient(); err != nil {
+		return nil, err
+	}
+
+	return k.client.Core().PersistentVolumes().Get(pvName, meta_v1.GetOptions{})
+}
+
+func (k *k8sOps) GetPersistentVolumes(namespace string) (*v1.PersistentVolumeList, error) {
```
piyush-nimbalkar: do we even need the namespace?

k8s/k8s.go
harsh-px: PVs aren't namespaced. So they are not needed. Removed namespace for PV functions.

PR #42, opened 2018-03-30
Add new interfaces to update deployments and statefulsets respectively

Rohit-PX:


k8s/k8s.go
```diff
@@ -930,6 +934,15 @@ func (k *k8sOps) DescribeDeployment(depName, depNamespace string) (*apps_api.Dep
 	return &dep.Status, err
 }
 
+func (k *k8sOps) UpdateDeployment(deployment *apps_api.Deployment) (*apps_api.Deployment, error) {
+	logrus.Info("Attempt to update the deployment.")
```
piyush-nimbalkar: nit: remove this log line.. if you want put in torpedo..

k8s/k8s.go
```diff
@@ -930,6 +934,15 @@ func (k *k8sOps) DescribeDeployment(depName, depNamespace string) (*apps_api.Dep
...
+	dep, err := k.appsClient().Deployments(deployment.Namespace).Update(deployment)
```
piyush-nimbalkar: initialize the client using initK8sClient() before you call the Update here.. refer GetDeployment()
This will fail if this is the first call and the client is not initialized

k8s/k8s.go
```diff
@@ -930,6 +934,15 @@ func (k *k8sOps) DescribeDeployment(depName, depNamespace string) (*apps_api.Dep
...
+	if err != nil {
```
piyush-nimbalkar: don't check error and return, just return the output as is
`return k.appsClient().Deployments(deployment.Namespace).Update(deployment)`
this is much simpler and the same thing that you are doing.

k8s/k8s.go
```diff
@@ -1365,6 +1378,14 @@ func (k *k8sOps) DescribeStatefulSet(ssetName string, ssetNamespace string) (*ap
 	return &sset.Status, err
 }
 
+func (k *k8sOps) UpdateStatefulSet(statefulset *apps_api.StatefulSet) (*apps_api.StatefulSet, error) {
+	ss, err := k.appsClient().StatefulSets(statefulset.Namespace).Update(statefulset)
```
piyush-nimbalkar: same comments as above

PR #46, opened 2018-04-18
Export api to get storage provisioner for PVC

harsh-px:


k8s/k8s.go
```diff
@@ -1989,6 +1991,15 @@ func (k *k8sOps) GetPVCsUsingStorageClass(scName string) ([]v1.PersistentVolumeC
 	return retList, nil
 }
 
+func (k *k8sOps) GetStorageProvisionerForPVC(pvc *v1.PersistentVolumeClaim) (string, error) {
```
piyush-nimbalkar: should we look at the provisioner annotation as well?

harsh-px: `getStorageClassForPV()` that gets called as part of this function does it.

harsh-px: Update PR to check for the provisioner annotation too.

PR #48, opened 2018-04-24
API to get and validate PVCs of stateful set

piyush-nimbalkar:


k8s/k8s.go
```diff
@@ -1926,11 +2000,15 @@ func (k *k8sOps) GetPersistentVolumeClaim(pvcName string, namespace string) (*v1
 }
 
 func (k *k8sOps) GetPersistentVolumeClaims(namespace string) (*v1.PersistentVolumeClaimList, error) {
+	return k.getPVCWithListOptions(namespace, meta_v1.ListOptions{})
+}
+
+func (k *k8sOps) getPVCWithListOptions(namespace string, listOpts meta_v1.ListOptions) (*v1.PersistentVolumeClaimList, error) {
```
adityadani: nit: getPVCsWithListOptions

k8s/k8s.go
```diff
@@ -1526,6 +1532,74 @@ func (k *k8sOps) GetStatefulSetsUsingStorageClass(scName string) ([]apps_api.Sta
 	return retList, nil
 }
 
+func (k *k8sOps) GetPVCForStatefulSet(name, namespace string) (*v1.PersistentVolumeClaimList, error) {
```
adityadani: nit: GetPVCsForStatefulSet

k8s/k8s.go
```diff
@@ -1526,6 +1532,74 @@ func (k *k8sOps) GetStatefulSetsUsingStorageClass(scName string) ([]apps_api.Sta
...
+
+	listOptions, err := k.getListOptionsForStatefulSet(ss)
+	if err != nil {
+		return nil, err
+	}
+
+	return k.getPVCWithListOptions(namespace, listOptions)
+}
+
+func (k *k8sOps) ValidatePVCForStatefulSet(name, namespace string) error {
```
harsh-px: duplicate code here. just call `k.GetPVCForStatefulSet` and then invoke `k.ValidatePersistentVolumeClaim`

`k.ValidatePersistentVolumeClaim` already has a retry loop for the PVC validation.

piyush-nimbalkar: We need the stateful set here to compare with PVC count with replica sets.

This retry also waits for the count of PVCs to match the replica sets.

PR #55, opened 2018-05-25
PWX-5477: added events op to create and list events

sdeoras:
This PR adds events op interface and implementation to create and list k8s events

Signed-off-by: Saurabh Deoras <sdeoras@gmail.com>

### DIFF
--- a/k8s/k8s.go
+++ b/k8s/k8s.go
@@ -70,13 +70,13 @@ type Ops interface {
 	SnapshotOps
 	SecretOps
 	ConfigMapOps
-	EventsOps
+	EventOps
 }
 
-// EventsOps is an interface to put and get k8s events
-type EventsOps interface {
+// EventOps is an interface to put and get k8s events
+type EventOps interface {
 	// CreateEvent puts an event into k8s etcd
-	CreateEvent(event *v1.Event) error
+	CreateEvent(event *v1.Event) (*v1.Event, error)
 	// ListEvents retrieves all events registered with kubernetes
 	ListEvents(namespace string, opts meta_v1.ListOptions) (*v1.EventList, error)
 }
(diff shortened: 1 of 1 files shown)


k8s/k8s.go
```diff
@@ -70,6 +70,15 @@ type Ops interface {
 	SnapshotOps
 	SecretOps
 	ConfigMapOps
+	EventsOps
```
piyush-nimbalkar: nit: EventOps to be coherent with other interfaces

k8s/k8s.go
```diff
@@ -70,6 +70,15 @@ type Ops interface {
...
+}
+
+// EventsOps is an interface to put and get k8s events
+type EventsOps interface {
+	// CreateEvent puts an event into k8s etcd
+	CreateEvent(event *v1.Event) error
```
harsh-px: Create should return the created event.

k8s/k8s.go
```diff
@@ -2471,6 +2480,17 @@ func (k *k8sOps) UpdateConfigMap(configMap *v1.ConfigMap) (*v1.ConfigMap, error)
 
 // ConfigMap APIs - END
 
+// CreateEvent puts an event into k8s etcd
+func (k *k8sOps) CreateEvent(event *v1.Event) error {
+	_, err := k.client.CoreV1().Events(event.Namespace).Create(event)
```
piyush-nimbalkar: initialize the client before the call.. 
```
if err := k.initK8sClient(); err != nil {
 		return err
  	}

k8s/k8s.go
sdeoras: done

k8s/k8s.go
sdeoras: done

k8s/k8s.go
sdeoras: done

PR #61, opened 2018-06-26
add GetPodByName

LynneD:
This function can get pod by pod name


k8s/k8s.go
```diff
@@ -1908,6 +1910,22 @@ func (k *k8sOps) listPluginPodsWithOptions(opts meta_v1.ListOptions, plugin stri
 	return retList, nil
 }
 
+func (k *k8sOps) GetPodByName(podName string, namespace string) (*v1.Pod, error) {
+	pods, err := k.GetPods(namespace)
```
disrani-px: Can get just the Pod instead of getting all pods and filtering it.

PR #65, opened 2018-07-10
Don't proceed with daemonset validation if observed generation is still 0

harsh-px:
0 observed generation would mean the daemonset controller hasn't even started monitoring the daemonset. https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.10/#daemonsetstatus-v1-apps

This was causing issues where other jobs like talisman declared daemonsets as ready even before they started deploying.

### DIFF
--- a/k8s/k8s.go
+++ b/k8s/k8s.go
@@ -1244,6 +1244,13 @@ func (k *k8sOps) ValidateDaemonSet(name, namespace string, timeout time.Duration
 			return "", true, err
 		}
 
+		if ds.Status.ObservedGeneration == 0 {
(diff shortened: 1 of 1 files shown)


k8s/k8s.go
```diff
@@ -1259,6 +1259,13 @@ func (k *k8sOps) ValidateDaemonSet(name, namespace string, timeout time.Duration
 			}
 		}
 
+		if ds.Status.ObservedGeneration == 0 {
```
adityadani: nit: this check should be before we call GetDaemonSetPods.

PR #70, opened 2018-07-22
Add interface to create CRD

harsh-px:
This moves the crd functions from https://github.com/portworx/talisman/tree/master/pkg/crd to here so other projects who are already vendoring sched-ops can use them

### DIFF
--- a/k8s/k8s.go
+++ b/k8s/k8s.go
@@ -24,7 +24,6 @@ import (
 	storage_api "k8s.io/api/storage/v1"
 	apiextensionsv1beta1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
 	apiextensionsclient "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
-	"k8s.io/apimachinery/pkg/api/errors"
 	meta_v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/selection"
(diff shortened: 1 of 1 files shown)


k8s/k8s.go
```diff
@@ -2665,76 +2696,143 @@ func (k *k8sOps) ListEvents(namespace string, opts meta_v1.ListOptions) (*v1.Eve
...
+				Singular: resource.Name,
+				Plural:   resource.Plural,
+				Kind:     resource.Kind,
+			},
+		},
+	}
+
+	_, err := k.apiExtensionClient.ApiextensionsV1beta1().CustomResourceDefinitions().Create(crd)
+	if err != nil {
+		if !errors.IsAlreadyExists(err) {
```
piyush-nimbalkar: I feel we should delegate the error handling to the caller. Sometimes it is better to know it is already created.

k8s/k8s.go
```diff
@@ -2665,76 +2696,143 @@ func (k *k8sOps) ListEvents(namespace string, opts meta_v1.ListOptions) (*v1.Eve
...
+	for _, resource := range resources {
+		if err := k.waitForCRDInit(resource); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (k *k8sOps) createCRD(resource CustomResource) error {
```
piyush-nimbalkar: We should expose a single CRD creation method as well and not just the list.

k8s/k8s.go
```diff
@@ -2665,76 +2696,143 @@ func (k *k8sOps) ListEvents(namespace string, opts meta_v1.ListOptions) (*v1.Eve
...
+func (k *k8sOps) CreateCRD(resources []CustomResource) error {
+	for _, resource := range resources {
+		err := k.createCRD(resource)
+		if err != nil {
+			return err
+		}
+	}
+
+	for _, resource := range resources {
+		if err := k.waitForCRDInit(resource); err != nil {
```
piyush-nimbalkar: We should have a separate method to validate/wait like we have for other objects instead of blocking the create call.

k8s/k8s.go
harsh-px: updated

k8s/k8s.go

k8s/k8s.go

PR #76, opened 2018-09-18
Add APIs for stork pairing and migration objects

disrani-px:


piyush-nimbalkar:
Now that we have stork client in the K8sOps, will it fail if stork is not running on a k8s cluster?
If yes, should we make storkclient optional or create a new StorkOps?

PR #77, opened 2018-09-24
Add List APIs for ClusterPair and Migration resources

disrani-px:
Signed-off-by: Dinesh Israni <disrani@portworx.com>


k8s/k8s.go
```diff
@@ -2752,6 +2756,14 @@ func (k *k8sOps) GetClusterPair(name string) (*v1alpha1.ClusterPair, error) {
 	return k.storkClient.Stork().ClusterPairs().Get(name, meta_v1.GetOptions{})
 }
 
+func (k *k8sOps) ListClusterPairs() (*v1alpha1.ClusterPairList, error) {
+	if err := k.initK8sClient(); err != nil {
+		return nil, nil
```
piyush-nimbalkar: return error here if there is an error
same for the other functions like create migration, create cluster pair, etc

PR #81, opened 2018-10-19
Remove vendor

harsh-px:
sched-ops is meant to be a pkg vendored by other repos. So it should not have it's vendor directory checked in. This can cause issues with nested vendoring.

One can run `dep ensure -vendor-only` in sched-ops to recreate the vendor directory.


disrani-px:
We should add `vendor` to `.gitignore` if we want to do this, otherwise `git status` will have too much noise when we pull in the dependencies

harsh-px:
> We should add vendor to .gitignore if we want to do this, otherwise git status will have too much noise when we pull in the dependencies


Done

PR #82, opened 2018-10-22
Allow to set new instance from kubeconfig file

ram-infrac:
This PR add support in k8s to return new k8s instance to be set. Will be required in case of multiple k8s client access.
-  It adds ValidateClusterpair() and VaildateMigration() api to check status

### DIFF
--- a/k8s/errors.go
+++ b/k8s/errors.go
@@ -3,6 +3,7 @@ package k8s
 import (
 	"errors"
 	"fmt"
+	"reflect"
 )
 
 // ErrK8SApiAccountNotSet is returned when the account used to talk to k8s api is not setup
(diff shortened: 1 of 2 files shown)


k8s/k8s.go
```diff
@@ -444,6 +446,8 @@ type MigrationOps interface {
 	UpdateMigration(*v1alpha1.Migration) (*v1alpha1.Migration, error)
 	// DeleteMigration deletes the Migration
 	DeleteMigration(string) error
+	// ValidateMigration validate the Migration status
+	ValidateMigration(name string, namespace string, timeout, retryInterval time.Duration)
```
disrani-px: Should return error. Travis is failing because of this.

k8s/k8s.go
```diff
@@ -2808,6 +2812,38 @@ func (k *k8sOps) DeleteClusterPair(name string) error {
...
+	if err := k.initK8sClient(); err != nil {
+		return err
+	}
+	t := func() (interface{}, bool, error) {
+		clusterPair, err := k.GetClusterPair(name)
+		if err != nil {
+			return "", true, err
+		}
+
+		if clusterPair.Status.SchedulerStatus == v1alpha1.ClusterPairStatusReady && clusterPair.Status.StorageStatus == v1alpha1.ClusterPairStatusReady {
```
disrani-px: Can you split this into multiple lines, doesn't even fit in the review page otherwise. Same below
```
if cond1 &&
   cond2 {
}

File: k8s/errors.go [hunk 3]
```diff
@@ -115,3 +115,15 @@ type ErrSnapshotDataFailed struct {
 func (e *ErrSnapshotDataFailed) Error() string {
 	return fmt.Sprintf("SnapshotData %v has failed. Cause: %v", e.ID, e.Cause)
 }
+
+// ErrFailedToApplyCustomSpec error type when CRD objects does not applied successfully
+type ErrFailedToApplyCustomSpec struct {
```
disrani-px: Should be `ErrFailedToValidateCustomSpec`, since the spec is already applied, but failing to reach Ready Status

disrani-px: Also add type to the structure so that we know what object is failing.

k8s/k8s.go
```diff
@@ -2854,6 +2890,38 @@ func (k *k8sOps) UpdateMigration(migration *v1alpha1.Migration) (*v1alpha1.Migra
 	return k.storkClient.Stork().Migrations().Update(migration)
 }
 
+func (k *k8sOps) ValidateMigration(name string, namespace string, timeout, retryInterval time.Duration) error {
+	if err := k.initK8sClient(); err != nil {
+		return err
+	}
+	t := func() (interface{}, bool, error) {
+		resp, err := k.GetMigration(name, namespace)
```
disrani-px: Migration doesn't have namespace

k8s/k8s.go
ram-infrac: looks like my entire commit is missing, sorry about this will update

PR #84, opened 2018-10-26
Bump vendor to kubernetes-1.12

harsh-px:
Had to upgrade github.com/kubernetes-incubator/external-storage from `v5.1.0` to `v5.2.0` since `v5.1.0` was using following client-go constraint (https://github.com/kubernetes-incubator/external-storage/blob/7613414913107241b76ab4db559386c3a04062b0/Gopkg.toml#L83)

```
[[constraint]]
  name = "k8s.io/client-go"
  version = "8.0.0"
```

This was conflicting the other repos that use client-go constraint

```
 [[constraint]]
   name = "k8s.io/client-go"
   version = "kubernetes-1.11.0"
```
or

```
 [[constraint]]
   name = "k8s.io/client-go"
   version = "kubernetes-1.12.0"
```

Also `dep ensure` was failing on sched-ops. Travis was previously passing since it was just using `dep ensure --vendor-only`. So it wasn't evaluating the Gopkg.toml again.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ go:
 install:
   - curl -s -L https://github.com/golang/dep/releases/download/v0.4.1/dep-linux-amd64 -o $GOPATH/bin/dep
   - chmod +x $GOPATH/bin/dep
-  - dep ensure -vendor-only
+  - dep ensure
   - go get -u github.com/vbatts/git-validation
   - go get -u golang.org/x/lint/golint
 script:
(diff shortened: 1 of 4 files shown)


harsh-px:
> This is going to cause issues when vendoring in to stork since all it's dependencies aren't at 1.12

We can bump stork packages to use 1.12 too. Or we can branch off a sched-ops kubernetes-1.11 branch from the previous commit if that's not possible.

I have to do this PR since `dep ensure` was failing with the existing Gopkg.toml.

PR #90, opened 2018-11-21
Update clusterpair and migration APIs to take namespace

disrani-px:


Gopkg.toml
```diff
@@ -42,6 +42,9 @@
   name = "k8s.io/client-go"
   version = "8.0.0"
 
+[[constraint]]
+  branch = "migration_scope"
```
disrani-px: Need to pin this to branch for now because fo circular dependencies. Will update to master once stork changes merge (which depends on this change)

PR #93, opened 2018-12-14
Use updated statefulset app received from GetStatefulSet()

ram-infrac:
Signed-off-by: Ram Suradkar <ram.suradkar@portworx.com>

### DIFF
--- a/k8s/k8s.go
+++ b/k8s/k8s.go
@@ -1609,7 +1609,7 @@ func (k *k8sOps) ValidateStatefulSet(statefulset *apps_api.StatefulSet, timeout 
 			return "", true, err
 		}
 
-		pods, err := k.GetStatefulSetPods(statefulset)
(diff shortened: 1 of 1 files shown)


k8s/k8s.go
```diff
@@ -1609,7 +1609,7 @@ func (k *k8sOps) ValidateStatefulSet(statefulset *apps_api.StatefulSet, timeout
 			return "", true, err
 		}
 
-		pods, err := k.GetStatefulSetPods(statefulset)
+		pods, err := k.GetStatefulSetPods(sset)
```
ram-infrac: is there any reason we were using **statefulset** instead of updated statefulset  recevied from **GetStatefulSet()** ?
For me UID of stateful set  in context of ClusterMigration required to be updated one

piyush-nimbalkar: I don't think there was any reason there, just a miss.
Good catch.

PR #98, opened 2019-01-09
 Add APIs for group volume snapshots

harsh-px:
PR adds sched-ops helper APIs for group volume snapshot CRDs


harsh-px:
Updated stork pkg to latest master.

PR #105, opened 2019-02-26
Add API to validate migration schedules

harsh-px:

### DIFF
--- a/k8s/k8s.go
+++ b/k8s/k8s.go
@@ -496,6 +496,12 @@ type MigrationOps interface {
 	ListMigrationSchedules(string) (*v1alpha1.MigrationScheduleList, error)
 	// DeleteMigrationSchedule deletes the MigrationSchedule
 	DeleteMigrationSchedule(string, string) error
+	// ValidateMigrationSchedule validates the given MigrationSchedule. It checks the status of each of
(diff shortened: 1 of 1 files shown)


k8s/k8s.go
```diff
@@ -3264,6 +3266,81 @@ func (k *k8sOps) DeleteMigrationSchedule(name string, namespace string) error {
...
+			return "", true, &ErrFailedToValidateCustomSpec{
+				Name:  name,
+				Cause: fmt.Sprintf("0 migrations have yet run for the migration schedule"),
+				Type:  resp,
+			}
+		}
+
+		failedMigrations := make([]string, 0)
+		pendingMigrations := make([]string, 0)
+		for _, migrationStatuses := range resp.Status.Items {
```
disrani-px: This can return false positives. If the caller wants to validate that a daily migration schedule has been triggered and there is only an interval schedule (just before the daily migration gets triggered) in successful state, it'll return a successful validation. The policy type to be validated also needs to be passed in, or return all the policy types that have a success and let the caller decide which one they want to check.

harsh-px: Updated the method to return a map of successful migrations. I'm still calling the method Validate as it does validate that atleast one migration is successful.

PR #111, opened 2019-03-27
Add create node and update pod APIs

harsh-px:
This is needed for fake client tests in stork unit tests

### DIFF
--- a/k8s/k8s.go
+++ b/k8s/k8s.go
@@ -114,6 +114,10 @@ type NamespaceOps interface {
 
 // NodeOps is an interface to perform k8s node operations
 type NodeOps interface {
+	// CreateNode creates the given node
+	CreateNode(n *v1.Node) (*v1.Node, error)
+	// UpdateNode updates the given node
+	UpdateNode(n *v1.Node) (*v1.Node, error)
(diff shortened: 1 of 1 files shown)


k8s/k8s.go
```diff
@@ -272,8 +274,9 @@ type RBACOps interface {
 
 // PodOps is an interface to perform k8s pod operations
 type PodOps interface {
-	// CreatePod creates the given pod
-	CreatePod(pod *v1.Pod) (*v1.Pod, error)
+	// CreatePod creates the given pod. If update is true, the API updates the given pod
+	// instead of creating a new pod
+	CreatePod(pod *v1.Pod, update bool) (*v1.Pod, error)
```
disrani-px: Every other API here has a separate UpdateObject API, we should keep it consistent

k8s/k8s.go
```diff
@@ -701,6 +707,21 @@ func (k *k8sOps) DeleteNamespace(name string) error {
 }
 
 // Namespace APIs - END
+func (k *k8sOps) CreateNode(n *v1.Node, update bool) (*v1.Node, error) {
```
harsh-px: need to fix this..hold on

PR #115, opened 2019-04-15
Add an API to validate cluster domains statuses

adityadani:


k8s/k8s.go
```diff
@@ -3827,6 +3830,49 @@ func (k *k8sOps) DeleteClusterDomainsStatus(name string) error {
 	})
 }
 
+func (k *k8sOps) ValidateClusterDomainsStatus(name string, domainMap map[string]bool, timeout, retryInterval time.Duration) error {
```
disrani-px: Don't you need a Validate for ClusterDomainUpdate also to make sure that it gets applied successfully?

adityadani: there is one already.

PR #122, opened 2019-05-10
Return detailed error when tasks time out

harsh-px:
Currently the method just returns "timed out performing task" and eats up the original error. 

Now returning the original error which will be useful in logs and also for clients to parse the error string.

Checked other repos and none currently is parsing the return type. So the API change should be okay.

### DIFF
--- a/task/task.go
+++ b/task/task.go
@@ -3,6 +3,7 @@ package task
 import (
 	"fmt"
 	"log"
+	"sync"
 	"time"
 )
 
@@ -22,10 +23,12 @@ func (e *ErrTimedOut) Error() string {
 func DoRetryWithTimeout(t func() (interface{}, bool, error), timeout, timeBeforeRetry time.Duration) (interface{}, error) {
 	done := make(chan bool, 1)
 	quit := make(chan bool, 1)
-	var out interface{}
-	var err error
-	var retry bool
-
+	var (
+		out     interface{}
+		err     error
+		errLock sync.Mutex
+		retry   bool
+	)
 	go func() {
 		count := 0
 		for {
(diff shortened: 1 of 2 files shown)


task/task.go
```diff
@@ -48,6 +55,8 @@ func DoRetryWithTimeout(t func() (interface{}, bool, error), timeout, timeBefore
 		return out, err
 	case <-time.After(timeout):
 		quit <- true
-		return out, ErrTimedOut
+		return out, &ErrTimedOut{
```
adityadani: Isn't there a race condition here as there is no lock over `err` object?
You might get a timeout, while err is being set to nil.

harsh-px: Good catch. Added a lock and also added UTs.

task/task.go
```diff
@@ -47,7 +59,11 @@ func DoRetryWithTimeout(t func() (interface{}, bool, error), timeout, timeBefore
 	case <-done:
 		return out, err
 	case <-time.After(timeout):
+		errLock.Lock()
+		defer errLock.Unlock()
 		quit <- true
-		return out, ErrTimedOut
+		return out, &ErrTimedOut{
+			Reason: err.Error(),
```
adityadani: Check if err is nil before calling err.Error()
There is still a possibility where the above goroutine succeeded and then released the lock and this code got it to return ErrTimedOut. 

To be safe, in addition to the above check for nil, we should do an errLock.Unlock() only after `true` is sent over the done channel.

task/task.go
```diff
@@ -29,7 +39,9 @@ func DoRetryWithTimeout(t func() (interface{}, bool, error), timeout, timeBefore
 				}
 
 			default:
+				errLock.Lock()
 				out, retry, err = t()
+				errLock.Unlock()
 				if err == nil || !retry {
 					done <- true
```
adityadani: errLock.Unlock() should be done after sending `true` over the done channel.

task/task.go
harsh-px: fixed

task/task.go
harsh-px: fixed

PR #123, opened 2019-05-14
Add APIs for ApplicationBackup, ApplicationRestore and BackupLocation

disrani-px:


k8s/k8s.go
```diff
@@ -4310,7 +4366,195 @@ func (k *k8sOps) UpdateObject(object runtime.Object) (runtime.Object, error) {
...
+
+func (k *k8sOps) GetApplicationBackup(name string, namespace string) (*v1alpha1.ApplicationBackup, error) {
+	if err := k.initK8sClient(); err != nil {
+		return nil, err
+	}
+
+	return k.storkClient.Stork().ApplicationBackups(namespace).Get(name, meta_v1.GetOptions{})
+}
+
+func (k *k8sOps) ListApplicationBackups(namespace string) (*v1alpha1.ApplicationBackupList, error) {
```
harsh-px: would be good to have the ListXXX API's take in list options.

disrani-px: We can add it later if required. Not all List APIs are passing that in now anyways.

PR #125, opened 2019-05-16
cleanup minor k8s version

harsh-px:
For git versions like v1.13.4+81fc896 the API has been found to return minor versions like 13+. This isn't very useful. So cleaning up minor to just have a number.


harsh-px:
This will be fixed in porx now: https://github.com/portworx/porx/pull/3889

PR #131, opened 2019-06-13
Add Validate APIs for BackupLocation and ApplicationBackup.

Rohit-PX:
Why do we need this PR?
Stork/Torpedo and tests would require these PRs for application backup testing.

### DIFF
--- a/k8s/k8s.go
+++ b/k8s/k8s.go
@@ -668,6 +668,8 @@ type BackupLocationOps interface {
 	UpdateBackupLocation(*v1alpha1.BackupLocation) (*v1alpha1.BackupLocation, error)
 	// DeleteBackupLocation deletes the BackupLocation
 	DeleteBackupLocation(string, string) error
+	// ValidateBackupLocation validates the BackupLocation
+	ValidateBackupLocation(string, string, time.Duration, time.Duration) error
 }
 
 // ApplicationBackupRestoreOps is an interface to perfrom k8s Application Backup
@@ -683,6 +685,8 @@ type ApplicationBackupRestoreOps interface {
 	UpdateApplicationBackup(*v1alpha1.ApplicationBackup) (*v1alpha1.ApplicationBackup, error)
 	// DeleteApplicationBackup deletes the ApplicationBackup
 	DeleteApplicationBackup(string, string) error
+	// ValidateApplicationBackup validates the ApplicationBackup
+	ValidateApplicationBackup(string, string, time.Duration, time.Duration) error
 	// CreateApplicationRestore creates the ApplicationRestore
 	CreateApplicationRestore(*v1alpha1.ApplicationRestore) (*v1alpha1.ApplicationRestore, error)
 	// GetApplicationRestore gets the ApplicationRestore
@@ -693,6 +697,8 @@ type ApplicationBackupRestoreOps interface {
 	UpdateApplicationRestore(*v1alpha1.ApplicationRestore) (*v1alpha1.ApplicationRestore, error)
 	// DeleteApplicationRestore deletes the ApplicationRestore
 	DeleteApplicationRestore(string, string) error
+	// ValidateApplicationRestore validates  the ApplicationRestore
+	ValidateApplicationRestore(string, string, time.Duration, time.Duration) error
 }
 
 // ApplicationCloneOps is an interface to perfrom k8s Application Clone operations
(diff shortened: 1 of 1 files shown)


k8s/k8s.go
```diff
@@ -4487,6 +4491,29 @@ func (k *k8sOps) UpdateBackupLocation(backupLocation *v1alpha1.BackupLocation) (
...
+func (k *k8sOps) ValidateBackupLocation(name, namespace string, timeout, retryInterval time.Duration) error {
+	t := func() (interface{}, bool, error) {
+		if err := k.initK8sClient(); err != nil {
+			return "", true, err
+		}
+
+		resp, err := k.GetBackupLocation(name, namespace)
+		if err != nil {
+			return "", true, err
+		}
```
harsh-px: instead of just checking we can do a get, we should also validate the response as the APIs intention is validate.

disrani-px: There is nothing to validate, there is no status for the backupLocation

k8s/k8s.go
```diff
@@ -4487,6 +4491,29 @@ func (k *k8sOps) UpdateBackupLocation(backupLocation *v1alpha1.BackupLocation) (
...
+		return "", true, &ErrFailedToValidateCustomSpec{
```
disrani-px: Need to return `"", false, nil` if err is nil. This will always return error right now.

k8s/k8s.go
```diff
@@ -4533,6 +4560,34 @@ func (k *k8sOps) UpdateApplicationBackup(backup *v1alpha1.ApplicationBackup) (*v
 	return k.storkClient.Stork().ApplicationBackups(backup.Namespace).Update(backup)
 }
 
+func (k *k8sOps) ValidateApplicationBackup(name, namespace string, timeout, retryInterval time.Duration) error {
```
disrani-px: Also need to add `ValidateApplicationRestore`

k8s/k8s.go
harsh-px: We can atleast check if `resp.Location` is populated.

disrani-px: This will just give false positives. Having anything set doesn't mean its valid.

k8s/k8s.go
```diff
@@ -4533,6 +4564,32 @@ func (k *k8sOps) UpdateApplicationBackup(backup *v1alpha1.ApplicationBackup) (*v
...
+	t := func() (interface{}, bool, error) {
+		if err := k.initK8sClient(); err != nil {
+			return "", true, err
+		}
+
+		applicationbackup, err := k.GetApplicationBackup(name, namespace)
```
disrani-px: Need to check for err and retry

k8s/k8s.go
```diff
@@ -4567,6 +4624,29 @@ func (k *k8sOps) DeleteApplicationRestore(name string, namespace string) error {
 	})
 }
 
+func (k *k8sOps) ValidateApplicationRestore(name, namespace string, timeout, retryInterval time.Duration) error {
+	t := func() (interface{}, bool, error) {
+		if err := k.initK8sClient(); err != nil {
+			return "", true, err
+		}
+
+		applicationrestore, err := k.storkClient.Stork().ApplicationRestores(namespace).Get(name, meta_v1.GetOptions{})
```
disrani-px: Need to check for err and retry

k8s/k8s.go
Rohit-PX: Fixed.

k8s/k8s.go
Rohit-PX: Fixed.

k8s/k8s.go
Rohit-PX: Fixed.

k8s/k8s.go
Rohit-PX: Fixed.

k8s/k8s.go
Rohit-PX: Fixed.

k8s/k8s.go
```diff
@@ -4567,6 +4625,32 @@ func (k *k8sOps) DeleteApplicationRestore(name string, namespace string) error {
...
+		if err != nil {
+			return "", true, nil
```
disrani-px: `return "", true, err`

k8s/k8s.go
```diff
@@ -4487,6 +4493,29 @@ func (k *k8sOps) UpdateBackupLocation(backupLocation *v1alpha1.BackupLocation) (
...
+	t := func() (interface{}, bool, error) {
+		if err := k.initK8sClient(); err != nil {
+			return "", true, err
+		}
+
+		resp, err := k.GetBackupLocation(name, namespace)
+		if err != nil {
+			return "", true, &ErrFailedToValidateCustomSpec{
+				Name:  name,
+				Cause: fmt.Sprintf("BackupLocation %v: %v failed . Error: %v", name, namespace, err),
```
disrani-px: Name is already being passed into the error, don't need to print it again.

Rohit-PX: Fixed.

k8s/k8s.go
Rohit-PX: Fixed.

PR #137, opened 2019-07-30
PTX-1709 Implementing Security Context Constaints for methods OC

nikolaypopov:
Implementing Security Context Constaints methods for OC:
Get
List
Update

Signed-off-by: nikolaypopov <nikolay.popov86@gmail.com>

### DIFF
--- a/k8s/k8s.go
+++ b/k8s/k8s.go
@@ -852,7 +852,7 @@ func (k *k8sOps) GetVersion() (*version.Info, error) {
 
 // Security Context Constraints APIs - BEGIN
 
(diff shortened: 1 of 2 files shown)


k8s/k8s.go
```diff
@@ -761,6 +776,7 @@ type k8sOps struct {
 	config             *rest.Config
 	dynamicInterface   dynamic.Interface
 	ocpClient          ocp_clientset.Interface
+	ocpSccClient       ocp_scc_clientset.Interface
```
piyush-nimbalkar: nit: Can we call this `ocpSecurityClient` instead of SccClient as it doesn't apply to only SCCs?

nikolaypopov: Done

PR #138, opened 2019-08-02
Add a method for registering CRD of kubernetes native type. Deprecate…

trierra:
…d the previous Create CRD method

### DIFF
--- a/k8s/k8s.go
+++ b/k8s/k8s.go
@@ -555,7 +555,10 @@ type ConfigMapOps interface {
 // CRDOps is an interface to perfrom k8s Customer Resource operations
 type CRDOps interface {
 	// CreateCRD creates the given custom resource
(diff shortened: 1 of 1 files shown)


harsh-px:
travis failing with

```
224  - FAIL - does not have a valid DCO
225  - FAIL - commit subject exceeds 90 characters.
```

k8s/k8s.go
```diff
@@ -554,8 +554,11 @@ type ConfigMapOps interface {
 
 // CRDOps is an interface to perfrom k8s Customer Resource operations
 type CRDOps interface {
-	// CreateCRD creates the given custom resource
+	// RegisterCRD creates the given custom resource
+	// Deprecated: Please use RegisterCRD instead
```
harsh-px: Change `This API will be deprecated soon. Use RegisterCRD instead`

PR #139, opened 2019-08-05
Add APIs for ApplicationBackupSchedule

disrani-px:
Signed-off-by: Dinesh Israni <disrani@portworx.com>

### DIFF
--- a/k8s/k8s.go
+++ b/k8s/k8s.go
@@ -719,6 +719,22 @@ type ApplicationBackupRestoreOps interface {
 	DeleteApplicationRestore(string, string) error
 	// ValidateApplicationRestore validates the ApplicationRestore
 	ValidateApplicationRestore(string, string, time.Duration, time.Duration) error
+	// GetApplicationBackupSchedule gets the ApplicationBackupSchedule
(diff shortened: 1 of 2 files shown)


k8s/k8s.go
```diff
@@ -4757,6 +4773,130 @@ func (k *k8sOps) UpdateApplicationRestore(restore *v1alpha1.ApplicationRestore)
...
+	if err := k.initK8sClient(); err != nil {
+		return err
+	}
+
+	return k.storkClient.Stork().ApplicationBackupSchedules(namespace).Delete(name, &meta_v1.DeleteOptions{
+		PropagationPolicy: &deleteForegroundPolicy,
+	})
+}
+
+func (k *k8sOps) ValidateApplicationBackupSchedule(name string, namespace string, timeout, retryInterval time.Duration) (
```
Rohit-PX: Validate should ideally return only error, instead of a map?

disrani-px: The caller also needs to validate that the backups that it expected are present. We can't know what the caller had expected here.

Rohit-PX: Sure, if that makes the tests simpler.

PR #144, opened 2019-08-21
Add AutopilotRule CRD CRUD operations

adityadani:
Signed-off-by: Aditya Dani <aditya@portworx.com>


k8s/k8s.go
```diff
@@ -89,6 +91,7 @@ type Ops interface {
 	ClusterPairOps
 	MigrationOps
 	ClusterDomainsOps
+	AutopilotRuleOps
```
harsh-px: The SetClient() might need to change. CI is failing due to that.

harsh-px: Pushed a fix now.

PR #151, opened 2019-09-17
Add empty import for "k8s.io/client-go/plugin/pkg/client/auth"

rodrigc:

### DIFF
--- a/k8s/k8s.go
+++ b/k8s/k8s.go
@@ -47,6 +47,7 @@ import (
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/kubernetes/scheme"
 	appsv1_client "k8s.io/client-go/kubernetes/typed/apps/v1"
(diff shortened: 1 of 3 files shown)


vendor/cloud.google.com/go/LICENSE
```diff
@@ -0,0 +1,202 @@
+
```
harsh-px: Don't checkin vendor directory. Libraries/packages should only declare their vendor requires in Gopkg.toml

rodrigc: OK, I removed the commit which updated the vendor tree

k8s/k8s.go
```diff
@@ -50,6 +50,8 @@ import (
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
 	"k8s.io/client-go/tools/remotecommand"
+
```
harsh-px: can u remove the empty line?

PR #157, opened 2019-10-02
Add SetConfigFromPath() which takes the path to a kubeconfig file

rodrigc:

### DIFF
--- a/k8s/k8s.go
+++ b/k8s/k8s.go
@@ -791,6 +791,8 @@ type ApplicationCloneOps interface {
 type ClientSetter interface {
 	// SetConfig sets the config and resets the client
 	SetConfig(config *rest.Config)
+	// SetConfigFromPath sets the config from a kubeconfig file
+	SetConfigFromPath(configPath string) error
 	// SetClient set the k8s clients
 	SetClient(
 		kubernetes.Interface,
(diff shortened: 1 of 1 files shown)


rodrigc:
This method is refactored from:
https://github.com/libopenstorage/stork/blob/master/test/integration_test/common_test.go#L251

k8s/k8s.go
```diff
@@ -858,6 +859,22 @@ func (k *k8sOps) SetConfig(config *rest.Config) {
...
+// and then internally calls SetConfig to set it
+func (k *k8sOps) SetRemoteConfig(kubeconfigPath string) error {
+	if kubeconfigPath == "" {
+		k.SetConfig(nil)
+		return nil
+	}
+	config, err := clientcmd.BuildConfigFromFlags("", kubeconfigPath)
+	if err != nil {
+		return err
+	}
```
piyush-nimbalkar: Why not just call this from your code and then call a SetConfig

rodrigc: @piyush-nimbalkar I wanted to see what people would think if we added this method directly to sched-ops.  Then any client of sched-ops, such as stork, spawn, torpedo, etc. could call this method, rather than having to re-implement the logic in each client.

harsh-px: Rename to `SetConfigFromPath`. That's more appropriate than remote.

harsh-px:
We can merge this once conflicts are resolved

piyush-nimbalkar:
Add this method under the new `ClientSetter` interface when resolving the conflicts

rodrigc:
I resolved the conflicts and added the method to the `ClientSetter` interface

PR #165, opened 2019-12-11
Remove init of auth plugins

disrani-px:
Projects that use auth plugins should init it there

### DIFF
--- a/k8s/k8s.go
+++ b/k8s/k8s.go
@@ -37,7 +37,6 @@ import (
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/kubernetes/scheme"
-	_ "k8s.io/client-go/plugin/pkg/client/auth" // needed to initialize plugins on cloud providers
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/remotecommand"
 )


disrani-px:
@rodrigc Looks like you had added this so you need to make sure to move this to where ever you were using the plugin

rodrigc:
I added this in https://github.com/portworx/sched-ops/pull/151/ which @harsh-px reviewed.
I did this when I integrated sched-ops into spawn.

It wasn't  obvious to me that a user of sched-ops is supposed to add this magic import in order for
integrating with cloud providers to work properly (AWS, Azure, etc.)

@harsh-px can you clarify this?  If this change is required for sched-ops, I'm OK with it as long as spawn and all other users of sched-ops change to support this.

harsh-px:
@disrani-px is this import causing a problem? if it's harmless, I would leave it here so cloud users of sched-ops don't have understand the internals of client-go on the need of this magic import.

The goal behind sched-ops is clients just create an instance and start using it.

disrani-px:
Yes it is causing an issue in another project because it is incorrect. It is bad practice to have anonymous imports anywhere but in the main package, static analysis tools will actually warn you about that.

Initializing plugins isn't the job of sched-ops, different users might want to use different implementations of the plugins.

harsh-px:
@rodrigc we can move this out of here and add it back in spawn and torpedo.

rodrigc:
@harsh-px it would be nice if we could have an explicit method for initializing the auth plugins instead of an anonymous import.  I don't know if that is doable or not, but from a end-user's perspective, it would make it more obvious as to what to do in order to initialize auth plugins for cloud providers.

harsh-px:
> It would be nice if we could have an explicit method for initializing the auth plugins instead of an anonymous import. I don't know if that is doable or not, but from a end-user's perspective, it would make it more obvious as to what to do in order to initialize auth plugins for cloud providers.

The init happens through an import. So can't do a method based approach. As Dinesh recommended, moving the import to the projects using sched-ops is a better practice.

PR #167, opened 2019-12-17
Changing Poll to PollImmediate to check condition before waiting for interval

piyush-nimbalkar:
Signed-off-by: Piyush Nimbalkar <piyush@portworx.com>


harsh-px:
Change looks harmless. Can you give some context on why this is needed?

piyush-nimbalkar:
I have some tests in the operator project that create and validate CRDs. Every test that goes through the validation takes at least 5 seconds which is the default interval in the actual code.

PR #173, opened 2020-01-10
[PWX-10495] Update project structure

saheienko:
Sched-ops has APIs for all the CRDs in one package. This introduces unnecessary dependencies for projects that want to vendor sched-ops and causes issues with version compatibilities.

The k8s package in sched-ops should be split up into different packages so that projects can vendor only the APIs they need and not the entire world.


k8s/client.go
```diff
@@ -2,6 +2,10 @@ package k8s
 
 import (
 	"fmt"
+	"github.com/portworx/sched-ops/k8s/autopilot"
+	"github.com/portworx/sched-ops/k8s/operator"
+	"github.com/portworx/sched-ops/k8s/stork"
+	"github.com/portworx/sched-ops/k8s/talisman"
```
stgleb: please apply goimports

k8s/autopilot/autopilot.go
```diff
@@ -0,0 +1,33 @@
+package autopilot
+
+import (
+	autopilotclientset "github.com/libopenstorage/autopilot-api/pkg/client/clientset/versioned"
+	"k8s.io/client-go/rest"
+)
+
+// AutopilotRuleOps provides an interface to Autopilot operations.
```
piyush-nimbalkar: I think it would be good to have a singleton instance of individual clients just like we have it for the k8sOps object. That way the users can just invoke the Instance anywhere without creating a new instance every time. 
```
import autopilotOps "github.com/portworx/sched-ops/k8s/autopilot"

autopilotOps.Instance().GetAutopilotRule(...)
```
We can do this lazily just like we did it before. In that case the individual implementations will look like this - 
```
func (c *Client) CreateAutopilotRule(rule *autv1alpaha1.AutopilotRule) (*autv1alpaha1.AutopilotRule, error) {
        if err := initAutopilotClient(); err != nil {
               return nil, err
        }
	return c.autopilot.AutopilotV1alpha1().AutopilotRules().Create(rule)
}
```
Refer `initK8sClient` method for above change.

k8s/autopilot/autopilot.go
```diff
@@ -0,0 +1,33 @@
...
+type AutopilotOps interface {
+	AutopilotRuleOps
+}
+
+func New(c autopilotclientset.Interface) *Client {
```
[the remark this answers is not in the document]
adityadani: +1

k8s/autopilot/autopilot.go
adityadani: +1

k8s/autopilot/autopilot.go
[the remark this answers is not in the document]
saheienko: That's not recommended to return an interface: https://github.com/golang/go/wiki/CodeReviewComments#interfaces
What the reason of doing this?

k8s/autopilot/autopilot.go
saheienko: Updated.

k8s/autopilot/autopilot.go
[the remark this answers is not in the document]
piyush-nimbalkar: This is interesting and good to know. Thanks for sharing. We have always been returning interfaces in New..() methods.
I am still wondering what good does it do to expose Client, especially here in this package where we have the interface as well as it's implementation. External consumers will mostly be assigning it to a Ops instead of Client instance.

k8s/autopilot/autopilot.go
```diff
@@ -0,0 +1,123 @@
...
+	return &Client{
+		autopilot: c,
+	}
+}
+
+// NewForConfig creates a new autopilot client for the given config.
+func NewForConfig(cfg *rest.Config) (*Client, error) {
```
piyush-nimbalkar: Actually do we need the New and NewForConfig method? It's basically an addition to the Instance() and SetConfig() + Instance() methods, which will return new instances instead of the singleton.

piyush-nimbalkar: Actually we will need a SetClient() for testing where we can pass a fake client to the singleton.

k8s/autopilot/autopilot.go
```diff
@@ -0,0 +1,123 @@
...
+import (
+	"fmt"
+	"os"
+
+	autopilotclientset "github.com/libopenstorage/autopilot-api/pkg/client/clientset/versioned"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
+)
+
+var (
```
stgleb: I am curious about that singleton. If you ask for instance it will have config and autopilot field null. 
Maybe you should initialize it with sync.Once to return configured client object.

piyush-nimbalkar: Good point. Should have something like this in the Instance() method.
```
func Instance() {
   once.Do(func() {
     instance = &Client{}
   })
   return instance
}
```

saheienko: @stgleb initialization is done lazily by initClient() method. [comment](https://github.com/portworx/sched-ops/pull/173#discussion_r365386365)

@piyush-nimbalkar there is no need for the once.Do functionc as instance object is defined not as a pointer and is initialized as a zero value.

stgleb: @saheienko I understand for instance as a value, but what about instance fields?

k8s/autopilot/autopilot.go
saheienko: New/NewForConfig returns a new instance of the client while Instance() returns the same one each time.

Would it be better to add a `SetInstance(client Ops)` function to the package (no need to add this to interface as it's for testing purpose)?

k8s/autopilot/autopilot.go
saheienko: @stgleb https://github.com/portworx/sched-ops/pull/173#discussion_r365386365
There is `initClient` invoked for each method that ensures client initialization. The same implementation approach is used for existing version of sched-ops.

PR #175, opened 2020-01-16
Cache modules in between travis builds

rodrigc:
This is similar to what was done here:
https://github.com/coredns/coredns/pull/2686/files#diff-354f30a63fb0907d4ad57269548329e3


harsh-px:
stale PR

PR #179, opened 2020-01-23
Keep k8s interface for backward compatibility

saheienko:
All the px projects are using k8s interface so it would be hard to migrate to specific clients at once. Keep k8s interface to allow these projects migrate gradually.


k8s/admissionregistration/admissionregistration.go
```diff
@@ -0,0 +1,137 @@
...
+	MutatingWebhookConfigurationOps
+
+	// SetConfig sets the config and resets the client.
+	SetConfig(config *rest.Config)
+}
+
+// Instance returns a singleton instance of the client.
+func Instance() Ops {
+	once.Do(func() {
+		if instance == nil {
```
piyush-nimbalkar: We don't need to do this nil check. `once.Do` ensure that the argument function is called only once in its lifetime.

k8s/client.go
```diff
@@ -0,0 +1,288 @@
...
+}
+
+// SetSnapshotClient sets the snapshot clientset
+func (k *k8sOps) SetSnapshotClient(snapClient rest.Interface) {
+	externalstorage.SetInstance(externalstorage.New(snapClient))
+}
+
+// SetStorkClient sets the stork clientset
+func (k *k8sOps) SetStorkClient(storkClient storkclientset.Interface) {
+}
```
piyush-nimbalkar: Missed set instance for stork ops.

k8s/client.go
```diff
@@ -0,0 +1,288 @@
...
+
+// SetDynamicClient sets the dynamic clientset
+func (k *k8sOps) SetDynamicClient(dynamicClient dynamicclient.Interface) {
+	dynamic.SetInstance(dynamic.New(dynamicClient))
+}
+
+// SetOpenshiftAppsClient sets the openshift apps clientset
+func (k *k8sOps) SetOpenshiftAppsClient(ocpAppsClient ocp_clientset.Interface) {
+}
```
piyush-nimbalkar: Missed set instance for openshift apps and security client below.

k8s/core/nodes.go
```diff
@@ -177,7 +177,7 @@ func (c *Client) SearchNodeByAddresses(addresses []string) (*corev1.Node, error)
 
 	// sweep #3 - locating based on labels
 	for _, node := range nodes.Items {
-		if hn, has := node.GetLabels()[corev1.LabelHostname]; has {
+		if hn, has := node.GetLabels()[labelHostname]; has {
```
piyush-nimbalkar: Why not use the label constant directly from kubernetes library?

k8s/k8s.go
```diff
@@ -0,0 +1,1291 @@
...
+	pvcStorageClassKey       = "volume.beta.kubernetes.io/storage-class"
+	pvcStorageProvisionerKey = "volume.beta.kubernetes.io/storage-provisioner"
+	labelUpdateMaxRetries    = 5
+)
+
+var deleteForegroundPolicy = meta_v1.DeletePropagationForeground
+
+// Ops is an interface to perform any kubernetes related operations
+type Ops interface {
+	NamespaceOps
```
piyush-nimbalkar: Instead of duplicating these individual ops, we should just compose this of the Ops that you extracted previously.
```
type Ops interface {
  core.Ops
  rbac.Ops
  ...
}
```
As the internal Ops have all of the below ops this list, we should directly be able to call, for instance `k8sOps.CreateServiceAccount()`

k8s/k8s.go
```diff
@@ -0,0 +1,1291 @@
...
+
+type privateMethods interface {
+}
+
+var (
+	instance Ops
+	once     sync.Once
+)
+
+type k8sOps struct{}
```
piyush-nimbalkar: This object just comprise of the extracted Ops objects. So when we return a new instance via `NewInstanceFromClients()` or `NewInstanceFromConfigFile()` it will be a brand new object with its own ops clients. We should not set the singleton instances of the individual ops in the `New..` methods.

k8s/client.go
```diff
@@ -0,0 +1,288 @@
...
+func (k *k8sOps) loadClientFromConfigBytes(kubeconfig []byte) error {
+	config, err := clientcmd.RESTConfigFromKubeConfig(kubeconfig)
+	if err != nil {
+		return err
+	}
+
+	return k.loadClientFor(config)
+}
+
+func (k *k8sOps) loadClientFor(config *rest.Config) error {
```
piyush-nimbalkar: I think we will not need any of the loadClient methods. Check comments in `k8s/k8s.go` for more details.

k8s/k8s.go
saheienko: Agree, removing these once.

k8s/admissionregistration/admissionregistration.go
saheienko: There is `SetInstance(i Ops)` func for each package that configures `instance`. So for such case `instance` value will be set to nil without the check:
```
i := fakeCoreClient{}
core.SetInstance(i)
core.Instance().ListPods() // here once.Do overrides the fake client
```

k8s/core/nodes.go
saheienko: This label has been added to the api package v1.14.x. This introduces an issues for projects which depend on older k8s packages.

k8s/k8s.go
```diff
@@ -0,0 +1,851 @@
...
+	dynamicInterface   dynamicclient.Interface
+	ocpClient          ocp_clientset.Interface
+	ocpSecurityClient  ocp_security_clientset.Interface
+	prometheusClient   prometheusclient.Interface
+}
+
+// Instance returns a singleton instance of k8sOps type
+func Instance() Ops {
+	once.Do(func() {
+		instance = &k8sOps{}
```
piyush-nimbalkar: From the above context, here we should just initialize all individual clients and not set the singleton instances of it.
```
instance = &k8sOps{
    CoreClient: &core.CoreClient{},
    StorkClient: &stork.StorkClient{},
    ...
}
```
I feel users should just avoid using the new and old way of using the k8s ops. It should be either the old way or the new split clients.

k8s/k8s.go
```diff
@@ -0,0 +1,851 @@
...
+	ClientSetter
+}
+
+var (
+	instance Ops
+	once     sync.Once
+)
+
+type k8sOps struct {
+	client             kubernetes.Interface
```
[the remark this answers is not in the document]
saheienko: Updated.

k8s/k8s.go
saheienko: Updated.

k8s/apiextensions/apiextensions.go
```diff
@@ -18,18 +18,25 @@ var (
...
+// Ops is an interface to the apiextensions client wrapper.
 type Ops interface {
 	CRDOps
 
 	// SetConfig sets the config and resets the client.
 	SetConfig(config *rest.Config)
 }
 
+// Interface is an interface to perform kubernetes related operations on the crd resources.
+type Interface interface {
```
adityadani: This interface is not being used anywhere. You need to use it in Ops

k8s/storage/volumeattachments.go
```diff
@@ -1,31 +1,31 @@
 package storage
 
 import (
-	storagev1 "k8s.io/api/storage/v1"
+	storagebeta "k8s.io/api/storage/v1beta1"
```
adityadani: why is this version change required? Or was this the original one before refactor?

saheienko: Yes, that's the original one.

k8s/apiextensions/apiextensions.go
saheienko: Updated.

saheienko:
Most of projects have been migrated to the new version of sched-ops. Closing it.

PR #182, opened 2020-01-28
adding NewInstanceFromConfigFile function

raviprasadsn:
Adding NewInstanceFromConfigFile api to return a new instance from given Kubeconfig file. This is needed for torpedo and px-backup.

### DIFF
--- a/k8s/admissionregistration/admissionregistration.go
+++ b/k8s/admissionregistration/admissionregistration.go
@@ -57,6 +57,17 @@ func NewForConfig(c *rest.Config) (*Client, error) {
 	}, nil
 }
 
+// NewInstanceFromConfigFile returns new instance of client by using given
+// config file
+func NewInstanceFromConfigFile(config string) (Ops, error) {
+	newInstance := &Client{}
+	err := newInstance.loadClientFromKubeconfig(config)
+	if err != nil {
+		return nil, err
+	}
+	return newInstance, nil
+}
+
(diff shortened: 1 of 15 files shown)


disrani-px:
@raviprasadsn do you need this for the other components too?

raviprasadsn:
> @raviprasadsn do you need this for the other components too?

Yes. Added it to all other components.

disrani-px:
@raviprasadsn Travis is failing because of missing DCO

raviprasadsn:
> @raviprasadsn Travis is failing because of missing DCO

Done. Signed-off

k8s/apiextentions/apiextentions.go
```diff
@@ -58,6 +59,18 @@ func NewForConfig(c *rest.Config) (*Client, error) {
 	}, nil
 }
 
+// NewInstanceFromConfigFile returns new instance of client by using given
+// config file
+func NewInstanceFromConfigFile(config string) (Ops, error) {
+	newInstance := &Client{}
+	err := newInstance.loadClientFromKubeconfig(config)
+	if err != nil {
+		logrus.Errorf("Unable to set new instance: %v", err)
```
disrani-px: I would skip the log message, we should avoid adding logs in packages that are used as libraries

raviprasadsn: Fixed.

raviprasadsn:
Added API to admission registration. Also added EventOps to CoreOps.

PR #183, opened 2020-01-29
PTX-2251 add support to  list services using filters

thiguetta:
Signed-off-by: Thiago Gonzaga <thi_gonzaga@yahoo.com.br>

### DIFF
--- a/k8s/core/services.go
+++ b/k8s/core/services.go
@@ -13,6 +13,10 @@ import (
 type ServiceOps interface {
 	// GetService gets the service by the name
 	GetService(string, string) (*corev1.Service, error)
+	// ListServices list services using filters or list all if options are empty
+	ListServices(string, metav1.ListOptions) (*corev1.ServiceList, error)
+	// GetServiceEndpoint gets the externalIP if service is a LoadBalancer or ClusterIP otherwise
+	GetServiceEndpoint(string, string) (string, error)
 	// CreateService creates the given service
 	CreateService(*corev1.Service) (*corev1.Service, error)
 	// DeleteService deletes the given service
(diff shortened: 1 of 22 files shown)


k8s/core/services.go
```diff
@@ -63,6 +65,15 @@ func (c *Client) GetService(svcName string, svcNS string) (*corev1.Service, erro
 	return c.core.Services(svcNS).Get(svcName, metav1.GetOptions{})
 }
 
+// ListServices list services using filters or list all if options are empty
+func (c *Client) ListServices(svcNS string, listOptions metav1.ListOptions) (*corev1.ServiceList, error) {
```
harsh-px: rename to `svcNamespace` to be consistent with below method

k8s/core/services.go
```diff
@@ -63,6 +67,31 @@ func (c *Client) GetService(svcName string, svcNS string) (*corev1.Service, erro
...
+func (c *Client) GetExternalIP(namespace string) (string, error) {
+	if err := c.initClient(); err != nil {
+		return "", err
+	}
+	svcs, err := c.ListServices(namespace, metav1.ListOptions{FieldSelector: "type=LoadBalancer"})
+	if err != nil {
+		return "", err
+	}
+	for _, svc := range svcs.Items {
+		return svc.Status.LoadBalancer.Ingress[0].IP, nil
```
nikolaypopov: So the first service out of the list of services will always have correct IP? And why are we always taking slice 0 of Ingress? Should we have some more asserts here? I'm not too familiar with, but is this enough to assume that first service with Ingress[0] will always have what we need?

k8s/core/services.go
```diff
@@ -13,6 +13,10 @@ import (
 type ServiceOps interface {
 	// GetService gets the service by the name
 	GetService(string, string) (*corev1.Service, error)
+	// ListServices list services using filters or list all if options are empty
+	ListServices(string, metav1.ListOptions) (*corev1.ServiceList, error)
+	// GetExternalIP
```
piyush-nimbalkar: nit: description missing

k8s/core/services.go
```diff
@@ -63,6 +67,31 @@ func (c *Client) GetService(svcName string, svcNS string) (*corev1.Service, erro
...
+	if err := c.initClient(); err != nil {
+		return nil, err
+	}
+
+	return c.core.Services(svcNamespace).List(listOptions)
+}
+
+// GetExternalIP gets the externalIP of a loadBalancer present in a given namespace, returns empty if LoadBalancer not
+// found
+func (c *Client) GetExternalIP(namespace string) (string, error) {
```
piyush-nimbalkar: Instead of just getting the external ip like this we should get service's endpoint. 
Basically given a service name and namespace, we should return cluster ip if the service type is ClusterIP (and if cluster ip is not none), ingress ip if the type is LoadBalancer. Let's leave node port out for now.

k8s/core/services.go
```diff
@@ -63,6 +67,30 @@ func (c *Client) GetService(svcName string, svcNS string) (*corev1.Service, erro
...
+func (c *Client) GetServiceEndpoint(svcName, namespace string) (string, error) {
+	if err := c.initClient(); err != nil {
+		return "", err
+	}
+	svc, err := c.GetService(svcName, namespace)
+	if err != nil {
+		return "", err
+	}
+	if svc.Spec.Type == corev1.ServiceTypeLoadBalancer {
+		return svc.Spec.LoadBalancerIP, nil
```
piyush-nimbalkar: I have seen most cloud providers use `svc.Status.LoadBalancer.Ingress.Hostname` or `svc.Status.LoadBalancer.Ingress.IP` and don't use the `spec.LoadBalancerIP`.
Let's first either of the above (if present) in this order - hostname -> ip -> spec.loadBalancerIP
If all are empty then return spec.ClusterIP itself.

PR #198, opened 2020-02-20
Fix for DoRetryWithTimeout when no retry with error

mborodin-px:
I'v noticed that if we use `return nil, false, err` in function which we pass to `DoRetryWithTimeout` it returns err as nil.

Signed-off-by: Maksym Borodin <maksym@portworx.com>

### DIFF
--- a/task/task.go
+++ b/task/task.go
@@ -48,7 +48,7 @@ func DoRetryWithTimeout(t func() (interface{}, bool, error), timeout, timeBefore
 				return
 			default:
 				out, retry, err := t()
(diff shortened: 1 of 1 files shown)


disrani-px:
@mborodin-px `TestDoRetryWithTimeoutSuccessNilReturn` needs to be fixed, the function should be returning `nil, false, nil`

mborodin-px:
@disrani-px I removed retry flag if it was no errors

disrani-px:
@mborodin-px Can you merge this? Some stork tests are failing because of this

PR #199, opened 2020-02-20
Use an explicit version of the stork client

saheienko:
```
// Deprecated: please explicitly pick a version if possible.
Stork() storkv1alpha1.StorkV1alpha1Interface
```
Use StorkV1Alpha1() instead of Stork() one.

Signed-off-by: Serhii Aheienko <serhii.aheienko@gmail.com>

### DIFF
--- a/k8s/stork/applicationbackuprestore.go
+++ b/k8s/stork/applicationbackuprestore.go
@@ -60,7 +60,7 @@ func (c *Client) CreateApplicationBackup(backup *storkv1alpha1.ApplicationBackup
 	if err := c.initClient(); err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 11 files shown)


disrani-px:
@saheienko Do you also need to run `make vendor` to update the go mod files?

saheienko:
There are no changes for deps as they are from the same interface:
```
type Interface interface {
	Discovery() discovery.DiscoveryInterface
	StorkV1alpha1() storkv1alpha1.StorkV1alpha1Interface
	// Deprecated: please explicitly pick a version if possible.
	Stork() storkv1alpha1.StorkV1alpha1Interface
}
```

PR #206, opened 2020-03-20
Add an EventRecorder API to sched-ops.

adityadani:
- The EventRecorder in client-go provides a way to record kubernetes events.
- In sched-ops we will maintain event recorders per unique event source.
- There will be a common event broadcaster for all the event recorders.

Signed-off-by: Aditya Dani <aditya@portworx.com>

### DIFF
--- a/k8s/core/events.go
+++ b/k8s/core/events.go
@@ -2,7 +2,12 @@ package core
 
 import (
 	corev1 "k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/kubernetes/scheme"
+	typedcorev1 "k8s.io/client-go/kubernetes/typed/core/v1"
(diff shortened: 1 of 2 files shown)


k8s/core/events.go
```diff
@@ -28,3 +33,33 @@ func (c *Client) ListEvents(namespace string, opts metav1.ListOptions) (*corev1.
...
+			},
+		)
+	}
+	eventRecorder, exists := c.eventRecorders[source.Component]
+	if !exists {
+		eventRecorder = c.eventBroadcaster.NewRecorder(
+			scheme.Scheme,
+			source,
+		)
+		c.eventRecorders[source.Component] = eventRecorder
```
piyush-nimbalkar: Should we use the key as combination of `source.Component + source.Host`? 
Not sure if we need it, but in case two sources have same component name but different host name, events might get created with the first source that comes here. 
Example I was thinking was, let's say we don't know the k8s node name and use node id but later want to use the k8s node name or the other way round.

adityadani: sounds good. Will change it to "source.Component-source.Host"

PR #214, opened 2020-04-17
Fix method name for Cron Job

Rohit-PX:


k8s/batch/cron.go
```diff
@@ -12,7 +12,7 @@ import (
 // Ops is an interface to perform kubernetes related operations on the crd resources.
 type CronOps interface {
 	// CreateCron creates the given cronJob
```
disrani-px, line 14: Comment needs update too

PR #216, opened 2020-05-06
Support for limitrange k8s object

Rohit-PX:
This is required for deploying and validating ingress controllers (specifically nginx ingress controllers)

### DIFF
--- a/k8s/core/limitrange.go
+++ b/k8s/core/limitrange.go
@@ -1,0 +1,90 @@
+package core
+
+import (
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/fields"
+)
+
+// LimitRangeOps is an interface to perform k8s LimitRange operations
+type LimitRangeOps interface {
+	// GetLimitRange gets the limitranges object given its name and namespace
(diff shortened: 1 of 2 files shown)


k8s/core/limitrange.go
```diff
@@ -0,0 +1,91 @@
+package core
+
+import (
+	corev1 "k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
```
saheienko, line 5: nit: `v1` and `corev1` name aliases point  to the same import.

k8s/core/limitrange.go
```diff
@@ -0,0 +1,91 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/fields"
+)
+
+// LimitRangeOps is an interface to perform k8s LimitRange operations
+type LimitRangeOps interface {
+	// GetLimitRange gets the secrets object given its name and namespace
```
saheienko, line 12: nit: replace `secrets object` with `limitrange object` for comments.

k8s/core/limitrange.go
Rohit-PX: fixed.

k8s/core/limitrange.go
Rohit-PX: fixed.

PR #219, opened 2020-07-06
add support to replicaset ops

thiguetta:
Signed-off-by: Thiago Gonzaga <thi_gonzaga@yahoo.com.br>

### DIFF
--- a/k8s/apps/replicasets.go
+++ b/k8s/apps/replicasets.go
@@ -1,0 +1,202 @@
+package apps
+
+import (
+	"fmt"
+	"time"
+
+	"github.com/portworx/sched-ops/k8s/common"
+	schederrors "github.com/portworx/sched-ops/k8s/errors"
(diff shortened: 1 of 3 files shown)


k8s/apps/replicasets.go
```diff
@@ -0,0 +1,202 @@
...
+		}
+
+		if rs.Status.ObservedGeneration == 0 {
+			return "", true, &schederrors.ErrAppNotReady{
+				ID:    name,
+				Cause: "Observed generation is still 0. Check back status after some time",
+			}
+		}
+
+		pods, err := c.GetReplicaSetPods(rs)
```
saheienko, line 95: Is retrieving the replicaset pods required here?
ReplicaSet controller updates rs status with pods information so it should be enough to check the status.

thiguetta: yep, it's needed since we need the pods to proper print the status of each pod if you check bellow

PR #237, opened 2021-02-11
Added support for node delete

prashanthpx:


k8s/core/nodes.go
```diff
@@ -55,6 +57,16 @@ func (c *Client) CreateNode(n *corev1.Node) (*corev1.Node, error) {
 	return c.kubernetes.CoreV1().Nodes().Create(n)
 }
 
+// DeleteNode deletes a node
```
ram-infrac, line 60: you will also need method to validate node delete for UT's, you can add to same PR or create new one

ram-infrac (approved):

siva-portworx (approved):

k8s/core/nodes.go
prashanthpx: Will add validate as separate PR.

PR #238, opened 2021-02-26
Upgrade to k8s 1.20 APIs

ggriffiths:
For an operator change, I need a k8s 1.18 client-go API change.

Upon updating to 1.20, I realized that there was a huge API change to add context to all client-go calls.

This means we need to update sched-ops to 1.20 k8s APIs because of Go's flattened dependency trees.

In other words, updating operator to k8s 1.20 requires sched-ops to use k8s 1.20, which requires all dependencies of sched-ops to use k8s 1.20.

Signed-off-by: Grant Griffiths <ggriffiths@purestorage.com>


harsh-px (approved):
Discussed offline with Grant.

```
Need to update sched-ops to k8s 1.20 for some operator CSI changes. In client-go 1.20, every client call requires a context object as the first argument. As a result, we need to update all client-go versions to k8s 1.20 (in autopilot, stork, etc).
Since there's a circulator test dependency, we may need to get the above sched-ops PR in first with failing tests. After that, we can update stork, operator, etc to pull in the latest sched-ops.
````

harsh-px:
will need for force push this since the CI will fail due to the cyclic dependency

ggriffiths:
Yep, I made a test tag `test-1.20` to test this update. We can force push once i test that w/ the operator code.

PR #242, opened 2021-03-24
Allow use of createOptions for apps resources

ram-infrac:


k8s/apps/deployments.go
```diff
@@ -21,7 +21,7 @@ type DeploymentOps interface {
 	// GetDeployment returns a deployment for the give name and namespace
 	GetDeployment(name, namespace string) (*appsv1.Deployment, error)
```
ram-infrac, line 22: Should we allow use of Get/update/delete options for all of sched-ops api as well

piyush-nimbalkar: Sure, we can add options to all APIs. 
The clients can update their code when they vendor the new changes.

piyush-nimbalkar (approved):

k8s/apps/deployments.go
ram-infrac: thanks,I will create separate PR

PR #249, opened 2021-07-27
build: bump k8s.io/client-go to v0.20.4

leseb:

This excludes pulling k8s.io/client-go(v12.0.0+incompatible) which
breaks integration with consumers of the
https://github.com/libopenstorage/secrets library since it pulls
https://github.com/portworx/sched-ops.

Also, sched-ops recently moved to k8s 1.20 in
https://github.com/portworx/sched-ops/pull/238 so I'm not sure why this
line was still around.

Signed-off-by: Sébastien Han <seb@redhat.com>

Thanks for considering this PR, if this is causing internal issues I have no control over then I can try to remove https://github.com/portworx/sched-ops as a dependency from https://github.com/libopenstorage/secrets since its usage is quite small.


leseb:
@harsh-px @disrani-px PTAL, thanks!

PR #251, opened 2021-08-23
add utils func to register v1 crds

ram-infrac:


piyush-nimbalkar:
Can we just use RegisterCRD function instead and also stop using the internal CustomResource object? Eventually we want everyone to use RegisterCRD with full schema while CreateCRD was meant to be deprecated.
If it's temporary, then I would say add a util in stork until then.

ram-infrac:
yes , it will be much cleaner if we do it via stork, I was not sure other repo apart from stork& operator using sched-ops lib to register crds?

piyush-nimbalkar:
> yes , it will be much cleaner if we do it via stork, I was not sure other repo apart from stork& operator using sched-ops lib to register crds?

Autopilot also uses it. We might have PDS use this in future.

ram-infrac:
so should we keep this util func ?

piyush-nimbalkar:
> so should we keep this util func ?

I think we should remove this. We are going to change autopilot too this month.

ram-infrac:
thanks piyush

PR #254, opened 2021-09-13
PB-1802 Added support to return all jobs in given namespace

prabhu-px:
Added support to return all jobs in given namespace
Closes #


k8s/batch/jobs.go
```diff
@@ -21,6 +21,8 @@ type JobOps interface {
 	// ValidateJob validates if the job with given namespace and name succeeds.
 	// It waits for timeout duration for job to succeed
 	ValidateJob(name, namespace string, timeout time.Duration) error
+	// GetAllJobs returns the jobs from given namespace
+	GetAllJobs(namespace string) (*batchv1.JobList, error)
```
siva-portworx, line 25: I think, it is better to rename it as ListJobs

prashanthpx (approved):

PR #261, opened 2021-10-01
PTX-3561 Torpedo reports 2/3 pods are ready, when all pods are ready

thiguetta:
Signed-off-by: Thiago Gonzaga <tgonzaga@purestorage.com>

This PR fixes an issue when kubelet is down in one node, pod is ready and running but ReadyReplicas field has a wrong value

### DIFF
--- a/k8s/apps/deployments.go
+++ b/k8s/apps/deployments.go
@@ -156,14 +156,6 @@ func (c *Client) ValidateDeployment(deployment *appsv1.Deployment, timeout, retr
 			}
 		}
 
-		if requiredReplicas > dep.Status.ReadyReplicas {
-			return "", true, &schederrors.ErrAppNotReady{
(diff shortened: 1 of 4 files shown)


k8s/apps/deployments.go
```diff
@@ -173,24 +165,22 @@ func (c *Client) ValidateDeployment(deployment *appsv1.Deployment, timeout, retr
...
-				notReadyPods = append(notReadyPods, pod.Name)
-			} else {
+			if common.IsPodReady(pod) {
 				readyCount++
 			}
 		}
 
-		if readyCount >= requiredReplicas {
-			return "", false, nil
+		if requiredReplicas > readyCount {
```
pault84, line 175: shouldn't this be `>=`

pault84 (changes requested):

PR #273, opened 2022-02-07
Adding DeleteStatefulSetPods func

nikolaypopov:
Adding DeleteStatefulSetPods func

Signed-off-by: nikolaypopov <nikolay.popov86@gmail.com>

### DIFF
--- a/k8s/apps/statefulsets.go
+++ b/k8s/apps/statefulsets.go
@@ -343,19 +343,12 @@ func (c *Client) DeleteStatefulSetPods(name, namespace string, timeout time.Dura
 		return err
 	}
 
-	var podsNamesToDelete []string
-	var podsToDelete []corev1.Pod
-	for _, pod := range pods {
-		podsNamesToDelete = append(podsNamesToDelete, pod.Name)
-		podsToDelete = append(podsToDelete, pod)
-	}
-
 	if err := common.DeletePods(c.core, pods, false); err != nil {
 		return err
 	}
 
-	if err := common.WaitForPodsToBeDeleted(c.core, podsToDelete, timeout); err != nil {
-		return fmt.Errorf("Failed to wait for pods to be deleted: %s, Err: %v", podsNamesToDelete, err)
+	if err := common.WaitForPodsToBeDeleted(c.core, pods, timeout); err != nil {
+		return fmt.Errorf("Failed to wait for pods to be deleted, Err: %v", err)
 	}
 
 	return nil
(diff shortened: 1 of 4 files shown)


k8s/common/pod.go
```diff
@@ -101,9 +101,9 @@ func WaitForPodsToBeDeleted(client v1.CoreV1Interface, podsToDelete []corev1.Pod
 }
 
 // WaitForPodDeletion waits for given timeout for given pod to be deleted
-func WaitForPodDeletion(client v1.CoreV1Interface, name string, namespace string, timeout time.Duration) error {
+func WaitForPodDeletion(client v1.CoreV1Interface, uid types.UID, namespace string, timeout time.Duration) error {
 	t := func() (interface{}, bool, error) {
-		p, err := GetPodByName(client, name, namespace)
+		p, err := GetPodByUID(client, uid, namespace)
```
piyush-nimbalkar, line 106: Instead of listing all the pods, can we just use GetPodByName and then verify that either the pod is not returned or does not have the passed UID. This will reduce the unnecessarily heavy call of getting all the pods in the system.

You can change the input to this function to just the get pod object.

k8s/apps/statefulsets.go
```diff
@@ -328,3 +330,33 @@ func (c *Client) validatePersistentVolumeClaim(pvc *corev1.PersistentVolumeClaim
...
+	pods, err := c.GetStatefulSetPods(sset)
+	if err != nil {
+		return err
+	}
+
+	var podsNamesToDelete []string
+	var podsToDelete []corev1.Pod
+	for _, pod := range pods {
+		podsNamesToDelete = append(podsNamesToDelete, pod.Name)
+		podsToDelete = append(podsToDelete, pod)
```
piyush-nimbalkar, line 350: Why can't we just use the list `pods` everywhere? Why do we need to construct this new list? It seems you are already using `pods` on DeletePods function but using `podsToDelete` in WaitForPodsToBeDeleted.

nikolaypopov: Done, removed it

k8s/common/pod.go
nikolaypopov: Done, added a check for comparing UIDs

k8s/common/pod.go
```diff
@@ -136,6 +136,23 @@ func GetPodByName(client v1.CoreV1Interface, podName string, namespace string) (
...
+	pUID := types.UID(uid)
+	for _, pod := range pods.Items {
+		if pod.UID == pUID {
+			return &pod, nil
+		}
+	}
+
+	return nil, schederrors.ErrPodsNotFound
+}
+
```
piyush-nimbalkar, line 155: We don't need this anymore.

piyush-nimbalkar (approved):
LGTM after removing the unused method.

k8s/common/pod.go
nikolaypopov: Removed

PR #276, opened 2022-03-08
Fixes ValidatePVCsForStatefulSet when apps have multiple PVCs in spec, and addresses flaky ConfigMap lock timing test

Pure-AdamuKaapan:
Currently if we try to use an app like `fio` which has both a logs and data PVC, this function will exit out of the count check early even if not all PVCs are created since it's only going by number of replicas.

Also fixes a configmap lock test that sometimes didn't wait long enough, probably just due to various factors over the 60 seconds.

Just finished testing this change in a torpedo deployment of mine, it counts correctly. ~~Will double check if there's a similar check anywhere else.~~ Just confirmed, this is the only time we check StatefulSet count.

### DIFF
--- a/k8s/apps/statefulsets.go
+++ b/k8s/apps/statefulsets.go
@@ -271,8 +271,9 @@ func (c *Client) ValidatePVCsForStatefulSet(ss *appsv1.StatefulSet, timeout, ret
 			return nil, true, err
 		}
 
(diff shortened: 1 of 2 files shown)


piyush-nimbalkar (approved):
Good catch and thanks for fixing it!

PR #280, opened 2022-03-23
[PB-2194]: backuplocation object to hold cluster credentials.

diptiranjanpx:
Signed-off-by: Diptiranjan <dsahoo@purestorage.com>

Backuplocation to have credentials needed to use cluster apis for taking snapshot and restore.

Closes # PB-2194


k8s/stork/backuplocation.go
```diff
@@ -68,6 +72,12 @@ func (c *Client) ListBackupLocations(namespace string, filterOptions metav1.List
 			return nil, err
 		}
 	}
+	for i := range backupLocations.Items {
+		err = backupLocations.Items[i].UpdateFromClusterSecret(c.kube)
```
prashanthpx, line 76: nit : As there is a for loop in the line#69, maybe this can be moved there itself.

prashanthpx (approved):

k8s/stork/backuplocation.go
diptiranjanpx: fixed.

PR #281, opened 2022-05-25
PWX-24105: check for nil value before dereferencing pod owner.Controller

pureneelesh:
OwnerReference.Controller is of type *bool. We were dereferencing that pointer
without checking if it was nil.

PWX-24105


jrivera-px (approved):
LGTM

PR #287, opened 2022-09-20
Allow users to set qps and burst rate for sched-ops api calls

ram-infrac:
This PR allows caller to set qps and burst rate for sched-ops api calls via env variables

Closes #

### DIFF
--- a/k8s/admissionregistration/admissionregistration.go
+++ b/k8s/admissionregistration/admissionregistration.go
@@ -5,6 +5,7 @@ import (
 	"os"
 	"sync"
 
+	"github.com/portworx/sched-ops/k8s/common"
 	apiadmissionsclientv1 "k8s.io/client-go/kubernetes/typed/admissionregistration/v1"
 	apiadmissionsclient "k8s.io/client-go/kubernetes/typed/admissionregistration/v1beta1"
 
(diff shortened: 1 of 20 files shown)


task/utils.go
```diff
@@ -0,0 +1,32 @@
...
+	"fmt"
+	"os"
+	"strconv"
+
+	"k8s.io/client-go/rest"
+)
+
+const (
+	QPSRate   = "QPS_RATE"
+	BurstRate = "BURST_RATE"
```
piyush-nimbalkar, line 13: Can we prefix the env vars so that it is specific to this library?
For instance, `KUBERNETES_OPS_QPS_RATE`?

task/utils.go
```diff
@@ -0,0 +1,32 @@
...
+)
+
+func SetRateLimiter(config *rest.Config) error {
```
piyush-nimbalkar, line 16: It doesn't seem right to put this in the `task` package as it is unrelated to k8s. We can use the `k8s/common` package.

piyush-nimbalkar (approved):
LGTM after comments

task/utils.go
ram-infrac: moved to k8s/common

PR #290, opened 2022-11-02
PWX-27654: CSR support

zoxpx:
Signed-off-by: Zoran Rajic <zrajic@purestorage.com>

This change adds support for Kubernetes CSR (Certificate Signing Requests).

Additionally, adding the `UpdateEvent()` to allow updating pre-existing K8s Events

Closes # PWX-27654

Required for "auto-SSL" setup of Portworx in K8s environments.

### DIFF
--- a/k8s/core/core.go
+++ b/k8s/core/core.go
@@ -9,6 +9,7 @@ import (
 	"github.com/portworx/sched-ops/k8s/common"
 	"github.com/portworx/sched-ops/task"
 	"github.com/sirupsen/logrus"
+	certv1 "k8s.io/api/certificates/v1"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -51,6 +52,7 @@ type Ops interface {
 	ServiceAccountOps
 	LimitRangeOps
 	NetworkPolicyOps
+	CertificateOps
 
 	// SetConfig sets the config and resets the client
 	SetConfig(config *rest.Config)
@@ -227,7 +229,8 @@ func (c *Client) handleWatch(
 	object runtime.Object,
 	namespace string,
 	fn WatchFunc,
-	listOptions metav1.ListOptions) {
+	listOptions metav1.ListOptions,
+) {
 	defer watchInterface.Stop()
 	for {
 		select {
@@ -245,6 +248,8 @@ func (c *Client) handleWatch(
 						err = c.WatchPods(namespace, fn, listOptions)
 					} else if sc, ok := object.(*corev1.Secret); ok {
 						err = c.WatchSecret(sc, fn)
+					} else if csr, ok := object.(*certv1.CertificateSigningRequest); ok {
+						err = c.WatchCertificateSigningRequests(csr, fn)
 					} else {
 						return "", false, fmt.Errorf("unsupported object: %v given to handle watch", object)
 					}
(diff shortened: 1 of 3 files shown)


k8s/core/core.go
```diff
@@ -9,6 +9,7 @@ import (
 	"github.com/portworx/sched-ops/k8s/common"
 	"github.com/portworx/sched-ops/task"
 	"github.com/sirupsen/logrus"
+	certv1 "k8s.io/api/certificates/v1"
```
piyush-nimbalkar, line 12: We have been creating separate clients under k8s folder for different API groups. Core, rbac, policy, etc.
Let's do the same for certificate APIs for consistency?

piyush-nimbalkar (approved):

k8s/core/core.go
zoxpx: > We have been creating separate clients under k8s folder for different API groups. Core, rbac, policy, etc. Let's do the same for certificate APIs for consistency?

@piyush-nimbalkar -- not sure what you are asking me to do here.  Can you give me an example?

Note that I had do add small changes into `core.go`, to include the CSR interface into the main sched-ops client, and also to handle the CSR watches.  The implementation went into a separate `csr.go` file.

zoxpx:
**UPDATE**: Sorry, didn't notice this before -- the code-comments were a bit off-sync.  Added follow-up [7e490ee](https://github.com/portworx/sched-ops/pull/290/commits/7e490eee5ab5f58044c072a446ef9f87706511e5) to fix.

k8s/core/core.go
piyush-nimbalkar: I meant move it to a separate package just like core and not just to a separate file - `k8s/certificates/csr.go`
We did that for all the different k8s apis.

zoxpx: > We did that for all the different k8s apis

I checked the [core.handleWatch()](https://github.com/portworx/sched-ops/blob/master/k8s/core/core.go#L225) -- and `WatchNode()`, `WatchConfigMap()`, `WatchPods()` and `WatchSecret()` are all implemented in the `core` package.
* since `Certificates` is IMO a concept very close to `Secrets` -- I followed how/where the `Secrets` API was implemented, and thus added it into the `core` package
* on the other hand, k8s-subdirs seem to be product-related  (`k8s/{stork,talisman,openshift,prometheus,autopilot}`...)

Is this going to be a problem?

zoxpx:
> **UPDATE**: Sorry, didn't notice this before -- the code-comments were a bit off-sync. Added follow-up [7e490ee](https://github.com/portworx/sched-ops/pull/290/commits/7e490eee5ab5f58044c072a446ef9f87706511e5) to fix.

Sorry, I forgot to add DSO to the commit, which broke Travis validation.   Fixed GIT comment + re-pushed the commit

k8s/core/core.go
piyush-nimbalkar: > k8s-subdirs seem to be product-related

That's not true. We have packages like rbac, apps, storage, batch, policy, dynamic, etc. All of them are separated by the Kubernetes API. Just because the concept is close to Secrets does not mean we should make an exception.

Of course, it is not going to a problem because this won't vendor new outside packages (one of the reasons why we separated these packages). It is just about consistency.

I have approved the PR already, so will leave it up to you. Just wanted to point out the inconsistency.

zoxpx:
Ok.. I have to make progress and start integrating this PR.   There's plenty more work that needs to follow up using this code (i.e. PX-Operator integration), so there will be opportunity to reorganize and optimize (also "battle-test") this code.

Thanks for the review Piyush

PR #294, opened 2022-12-22
PWX-28162_pt2: Add CSR update API call

zoxpx:

Adding a call to update approval-status on CertificateSigningRequest (CSR) objects
* this call is needed by the PX-Operator, to automatically approve the CSR objects posted by the OCI-Monitor

Closes #  PWX-28162 (part 2)

### DIFF
--- a/k8s/core/csr.go
+++ b/k8s/core/csr.go
@@ -158,6 +158,7 @@ func (c *Client) WatchCertificateSigningRequests(csr *certv1.CertificateSigningR
 	return nil
 }
 
+// CertificateSigningRequestsUpdateApproval used to approve or decline the CSR
 func (c *Client) CertificateSigningRequestsUpdateApproval(
 	name string,
 	csr *certv1.CertificateSigningRequest,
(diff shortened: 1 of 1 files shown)


jrivera-px (approved):
How is this call used with the "cluster-id" from your previous PR?

piyush-nimbalkar (approved):

zoxpx:
UPDATE: With [6b33300](https://github.com/portworx/sched-ops/pull/294/commits/6b333003acbb22e9d69b9aed9317c894e2dd4f8c) fixed commit-comment to add a DCO  -- since travis was not happy about this

zoxpx:
> How is this call used with the "cluster-id" from your previous PR?

This fix is not related to adding "ClusterID" into CSR MetaData.

* we added ClusterID into the CSR Metadata, so that the PX-Operator has better information which CSR's should get auto-approved, or ignored
  - also, PX-Operator needs to respect "disable auto-approval" flag set on StorageCluster, which is ClusterID-specific setting
* this change is required to actually Approve (or Deny) the posted CSR requests
  - so, after PX-Operator figures out it will auto-approve the CSR, it will use this "update" API to do so

The "third part" of the fix, is actually the code in PX-Operator that automatically approves the CSRs posted by OCI-Monitor.

zoxpx:
Thanks for the review Jose and Piyush -- merging the PR.

PR #297, opened 2023-02-05
pb-3338: Added GetDefaultStorageClasses api to get the list of default storageclass.

siva-portworx:
```
 pb-3338: Added GetDefaultStorageClasses api to get the list of default storageclass
```
Closes # pb-3338

Tested the API with stork code.

### DIFF
--- a/k8s/storage/storageclass.go
+++ b/k8s/storage/storageclass.go
@@ -14,6 +14,8 @@ type ScOps interface {
 	GetStorageClasses(labelSelector map[string]string) (*storagev1.StorageClassList, error)
 	// GetStorageClass returns the storage class for the give namme
 	GetStorageClass(name string) (*storagev1.StorageClass, error)
+	// GetDefaultStorageClasses returns all storageClasses that are set as default
+	GetDefaultStorageClasses() (*storagev1.StorageClassList, error)
 	// CreateStorageClass creates the given storage class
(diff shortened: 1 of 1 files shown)


k8s/storage/storageclass.go
```diff
@@ -46,6 +50,28 @@ func (c *Client) GetStorageClass(name string) (*storagev1.StorageClass, error) {
...
+        if err := c.initClient(); err != nil {
+                return nil, err
+        }
+
+        storageClassList, err := c.storage.StorageClasses().List(context.TODO(), metav1.ListOptions{})
+        if err != nil {
+                return nil, err
+        }
+        for _, sc := range storageClassList.Items {
+                if val, ok := sc.Annotations[defaultStorageclassAnnotationKey]; ok {
```
prashanthpx, line 66: So here if they have annotations set on different storage classes, then we would gibe out the first one from the is it? 
Everytime this API is called then there is no guarantee it would be same SC in case of multiple default SC is set?

prashanthpx (commented):
lgtm

k8s/storage/storageclass.go
siva-portworx: Yes, sched-ops will return all the storageclass that are set to default. If more than one sc is set as default, it is up to callers to handle the return value.
In the case of stork, I am handling both the number of default sc greater than one or zero.

diptiranjanpx (approved):

PR #302, opened 2023-02-17
PWX-26175_r2: Adding RunCommandInPodEx

zoxpx:
* adding `RunCommandInPodEx()`, as "extended" version of func that runs arbitrary commands in a POD
  - this version can pass STDIN to the command, also captures STDOUT/STDERR as separate outputs
* the original `RunCommandInPod()` still exists with same params -- but was refactored to use the new "extended" `RunCommandInPodEx()` internally

Signed-off-by: Zoran Rajic <zrajic@purestorage.com>

Needed for integration-tests added into PX-Operator  (see https://github.com/libopenstorage/operator/pull/938)

Closes #  PWX-26175  (part 2)

See also [operator#938](https://github.com/libopenstorage/operator/pull/938) where we use this func.

### DIFF
--- a/k8s/core/pods.go
+++ b/k8s/core/pods.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"os"
 	"time"
 
 	"github.com/portworx/sched-ops/k8s/common"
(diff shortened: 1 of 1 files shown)


k8s/core/pods.go
```diff
@@ -69,6 +69,8 @@ type PodOps interface {
 	WaitForPodDeletion(uid types.UID, namespace string, timeout time.Duration) error
 	// RunCommandInPod runs given command in the given pod
 	RunCommandInPod(cmds []string, podName, containerName, namespace string) (string, error)
+	// RunCommandInPodEx is extended version of RunCommandInPod
+	RunCommandInPodEx(cmds []string, podName, containerName, namespace string, useTTY bool, stdin io.Reader, stdout, stderr io.Writer) error
```
ggriffiths, line 73: I see we are adding additional parameters here and changing the interface function name.

When we update this to add a new parameter, it'll have to yet another new name to avoid breaking changes.

A better approach for this would be to use request object - i.e. 
```
type RunCommandInPodRequest struct {
  Cmds []string
  podName string
  ...
  stdout io.Writer
  stderr io.Writer
}
```

zoxpx: > I see we are adding additional parameters here and changing the interface function name.

Yes -- the original `RunCommandInPod()` did not take the STDIN param, also did not separate STDOUT and STDERR streams.  The "extended" `RunCommandInPodEx()` does all this.

> A better approach for this would be to use request object - i.e.

So.. you're just asking to change the 8 input parameters, and take only the 1 param (the `RunCommandInPodRequest struct`)?   Note that this will "break" the uniformity of the API  (8 input params is "a lot", good point there, but no other POD-function takes struct as an input param).

ggriffiths: I would suggest one new function RunCommandInPodEx with a request object `RunCommandInPodExRequest`.

That way, you can add parameters without changing the function signature every time. Makes it more maintainable

zoxpx: Ok.. change implemented in [dfcfbc0](https://github.com/portworx/sched-ops/pull/302/commits/dfcfbc024ba7381a445de458588063f067de59a8)

ggriffiths (approved):
lgtm thanks @zoxpx

zoxpx:
Thanks Grant -- merging the PR

PR #304, opened 2023-03-13
PWX-29473: add a function to allow a different lock hold timeout

pureneelesh:
Add a LockWithHoldTimeout function to allow caller to specify a different lock hold timeout than what was specified during object creation.

PWX-29473


zoxpx (approved):
Looks harmless enough -- thanks Neelesh

PR #314, opened 2023-04-28
PWX-30230 : Create all px configmaps in same namespace as portworx

nikita-bhatia:
https://portworx.atlassian.net/browse/PWX-30230


k8s/core/configmap/configmap_test.go
```diff
@@ -17,7 +17,8 @@ func TestGetConfigMap(t *testing.T) {
 	configData := map[string]string{
 		"key1": "val1",
 	}
-	cm, err := New("px-configmaps-test", configData, lockTimeout, 5, 0, 0)
+	cm, err := New("px-configmaps-test", configData, lockTimeout, 5, 0, 0, "test-ns")
+	fmt.Println("cm : ", cm)
```
nrevanna, line 21: was this for dev debugging? Remove?

k8s/core/configmap/configmap.go
```diff
@@ -1,188 +1,101 @@
...
 
-	if v2LockK8sLockTTL == 0 {
-		v2LockK8sLockTTL = v2DefaultK8sLockTTL
+	lockMap, err := c.copylock.get()
+	if err != nil {
+		log.Errorf("Error during fetching data from copy lock %s", err)
+		return existingConfig
 	}
+	status := lockMap[upgradeCompletedStatus]
+	if status == trueString {
```
nrevanna, line 27: How are we handling IN_PROGRESS state.
Let's say N1 is doing the copy. At the same time N2 is trying to update some value usingUpdate() function. According to this implementation, on N2 existingConfig will be returned and could be updating while N1 could be deleting the whole configmap at line 52. Don't we need to wait if a copy is IN_PROGRESS?

nrevanna:
Can you add testing information. Also share a testplan. This one will need to tested well. Would prefer to have it all documented in testrail.

PR #320, opened 2023-07-03
Cherrypick ns

nikita-bhatia:
https://portworx.atlassian.net/browse/PWX-30230

This PR is for migration of namespace from 'kube-system' to 'portworx'. To support this,

Added coreConfigMap struct as wrapper function for existing configmap.
Added Instance() to decide which configmap to use, and handle upgrade scenarios
Logic to update copylock config map entry once upgrade is finished will be handled in porx code.
This PR is to cherrypick changes to 'porx' branch of sched-ops, to update vendor dependency in porx repo.

### DIFF
--- a/k8s/core/configmap/configmap.go
+++ b/k8s/core/configmap/configmap.go
@@ -7,26 +7,23 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
-func (c *configMap) Instance() ConfigMap {
-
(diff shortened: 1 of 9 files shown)


nrevanna (commented):
This one seems to be similar to https://github.com/portworx/sched-ops/pull/314 except for configmap_test. Why do we have two PRs?

PR #322, opened 2023-08-17
PWX-32793 : add support for raw block devices

ashinde-px:
Add support for raw block devices
Closes #PWX-32793

- In listing pods using pvc we were checking volume mounts to list the pods
- In case of raw block volumes, we need to check volume devices instead of volume mounts, hence we need to update this method for listing pods using pvc's


sharma-tapas (approved):

Pure-AdamuKaapan (approved):
LGTM - if we can add some unit tests that would be good but if there aren't existing tests don't bother.

PR #324, opened 2023-09-12
PWX-33673 Added check before unlock

rkinhalkar-px:
Added check before unlock

PWX-33673

### DIFF
--- a/k8s/core/configmap/configmap_lock_v1.go
+++ b/k8s/core/configmap/configmap_lock_v1.go
@@ -44,7 +44,12 @@ func (c *configMap) Unlock() error {
 	fn := "Unlock"
 	// Get the existing ConfigMap
 	c.kLockV1.Lock()
-	defer c.kLockV1.Unlock()
+	defer func() {
+		// Check if the mutex is locked before unlocking
+		if !c.kLockV1.unlocked {
+			c.kLockV1.Unlock()
+		}
+	}()
 	if c.kLockV1.unlocked {
 		// The lock is already unlocked
 		return nil


k8s/core/configmap/configmap_lock_v1.go
```diff
@@ -44,7 +44,12 @@ func (c *configMap) Unlock() error {
 	fn := "Unlock"
 	// Get the existing ConfigMap
 	c.kLockV1.Lock()
-	defer c.kLockV1.Unlock()
+	defer func() {
+		// Check if the mutex is locked before unlocking
+		if !c.kLockV1.unlocked {
+			c.kLockV1.Unlock()
+		}
+	}()
```
nrevanna, line 52: unlocked is set to true always in line 57. We will not unlock this every in that case. I don't think that should be the fix here. We need to see where c.KLockV1's Unlock is happening more than once

pnookala-px: The only thing I can think about is that klockv1 changed after we did the lock because of which it is unlocked state. Is it possible that klockv1 during the life of the cm.

harsh-px (changes requested):
PR lacks testing notes. Please add.

k8s/core/configmap/configmap_lock_v1.go
nrevanna: The other possibility is that some other path did an unlock unnecessarily. 
I had a chat with Rushi. He said he is able to reproduce the issue. I have suggested to trace all locks and unlocks and find an unexpected unlock that might have occured.

rkinhalkar-px:
closing this PR,
creating a new one with new changes

PR #327, opened 2023-09-22
add a dynamic client for KubeVirt

pureneelesh:

The client currently returns only a few selected fields from the VirtualMachineInstance CR.

Closes #

Verified by manually running the unit test on the live cluster.

### DIFF
--- a/k8s/kubevirt-dynamic/kubevirt.go
+++ b/k8s/kubevirt-dynamic/kubevirt.go
@@ -1,0 +1,224 @@
+package kubevirtdynamic
+
+import (
+	"fmt"
+	"os"
+	"sync"
+
+	"github.com/portworx/sched-ops/k8s/common"
+	"k8s.io/client-go/dynamic"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
+)
+
+var (
+	instance Ops
+	once     sync.Once
+)
+
+// Ops is an interface to perform generic Object operations
+type Ops interface {
+	VirtualMachineInstanceOps
+
+	// SetConfig sets the config and resets the client
+	SetConfig(config *rest.Config)
+}
+
+// Instance returns a singleton instance of the client.
+func Instance() Ops {
+	once.Do(func() {
+		if instance == nil {
+			instance = &Client{}
+		}
+	})
+	return instance
+}
+
+// SetInstance replaces the instance with the provided one. Should be used only
+// for testing purposes.
+func SetInstance(i Ops) {
+	instance = i
+}
+
+// New builds a new client.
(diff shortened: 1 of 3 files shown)


k8s/kubevirt/kubevirt.go
```diff
@@ -0,0 +1,366 @@
...
+	conditions, found, err := unstructured.NestedSlice(vmiRaw.Object, "status", "conditions")
+	if err != nil {
+		return nil, fmt.Errorf("failed to find conditions in vmi: %w", err)
+	}
+	if found {
+		condition, err := c.unstructuredFindKeyValString(conditions, "type", "LiveMigratable")
+		if err != nil {
+			return nil, fmt.Errorf("failed while finding live migratable condition in vmi: %w", err)
+		}
+		if condition != nil {
```
pp511, line 209: nit.. Maybe we can return err when "condition" is found to be nil. Just like we are returning in case of rootVolume == nil above. I agree that If LiveMigratable condition is not present, we should consider it as false. However we can leave that to the caller to decide and let this function report error. Let me know if you disagree.

pp511 (commented):
Changes look good.
Just one nit I need to hear your opinion on.

k8s/kubevirt/kubevirt.go
pureneelesh: Discussed offline. We will treat the missing LiveMigratable condition same as LiveMigratable=false for now since it satisfies our current use cases. When we move to the typed client, we can return the original VMI as it is and let the caller decide.

pp511 (approved):

k8s/kubevirt/kubevirt.go
```diff
@@ -0,0 +1,366 @@
...
+}
+
+// Ops is an interface to perform generic Object operations
+type Ops interface {
+	// GetVirtualMachineInstance retrieves some info about the specified VMI
+	GetVirtualMachineInstance(ctx context.Context, namespace, name string) (*VirtualMachineInstance, error)
+
+	// SetConfig sets the config and resets the client
+	SetConfig(config *rest.Config)
+}
```
piyush-nimbalkar, line 41: To follow the same pattern as other schedops packages, can we create a separate interface VMI operations and add that interface in the parent `Ops` interface? Also the VMI related code can be moved to a separate file called `vmi.go` or `virtualmachineinstance.go`.

```
> kubevirt.go

type Ops interface {
    VirtualMachineInstanceOps
    SetConfig(...)
}
```
```
> virtualmachineinstance.go

type VirtualMachineInstanceOps {
    GetVirtualMachineInstance(...)
}
```

piyush-nimbalkar (approved):

PR #328, opened 2023-09-22
PTX-20287 Adding method to set config from file for Operator Marketplace APIs

nikolaypopov:
Adding method to set config from file for Operator Marketplace APIs


nikolaypopov:
Not needed anymore

PR #336, opened 2023-10-13
PB-4601: Upon restoring kubevirt VM, there is a change in VM spec par…

vsundarraj-px:
…ameter

Fixes PB-4601. Not all VMs have spec.Running. Stork removes Spec.Running element and replaces it with runStrategy. This causes validateVirtualMachineRunningState fails. This PR fixes this to handle it gracefully. 
Closes #
PB-4601
None

### DIFF
--- a/k8s/kubevirt/virtualmachine.go
+++ b/k8s/kubevirt/virtualmachine.go
@@ -76,8 +76,9 @@ func (c *Client) ValidateVirtualMachineRunning(name, namespace string, timeout, 
 	}
 
 	// Start the VirtualMachine if its not Started yet
-	if !*vm.Spec.Running {
-		if err = instance.StartVirtualMachine(vm); err != nil {
+	if vm.Status.PrintableStatus == kubevirtv1.VirtualMachineStatusStopped ||
+		vm.Status.PrintableStatus == kubevirtv1.VirtualMachineStatusStopping {
+		if err = c.StartVirtualMachine(vm); err != nil {
 			return fmt.Errorf("Failed to start VirtualMachine %v", err)
 		}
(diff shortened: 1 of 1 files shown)


k8s/kubevirt/virtualmachine.go
```diff
@@ -76,8 +76,9 @@ func (c *Client) ValidateVirtualMachineRunning(name, namespace string, timeout,
...
 
 	// Start the VirtualMachine if its not Started yet
-	if !*vm.Spec.Running {
-		if err = instance.StartVirtualMachine(vm); err != nil {
+	if vm.Status.PrintableStatus == kubevirtv1.VirtualMachineStatusStopped ||
+	   vm.Status.PrintableStatus == kubevirtv1.VirtualMachineStatusStopping {
+		if err = c.StartVirtualMachine(vm); err != nil {
 			return fmt.Errorf("Failed to start VirtualMachine %v", err)
 		}
 	}
```
mkoppal-px, line 84: Now that I think of this, do we need to start the VM if it is stopped?
We have cases where we expect the VMs to be in stopped state, but this is going to start it. Here we should just be checking if it is running right?

mkoppal-px (approved):
LGTM

k8s/kubevirt/virtualmachine.go
vsundarraj-px: As discussed this function is intended to verify start soon after the VM is created. This will start if the creatVM does not start the VM automatically and wait for VM to start. 
For generic start/stopvm we should expose other method isRunning(vm) and use Start/Stop method which is already exposed.

PR #343, opened 2023-11-21
PB-4403: alert manager config for email alert

t-nidhin:
- Email alert requires alertmanagerconfig to create receivers list, which then gets picked up by alert manager to send emails based on respective logs.

- This is the api px-backup uses to manage the amc

Closes #
This is based on the ticket of [pb-4403](https://portworx.atlassian.net/browse/PB-4403)


px-kesavan (approved):
fix travis

PR #346, opened 2023-11-29
PB-4953: Kubevirt - Phase 2 Implement Rule Add for freeze/thaw

vsundarraj-px:
This is sched-ops implementation for Kubevirt phase 2. This new APIs will be used by PXB and Torpedo

Closes #PB-4953

Unit testing is done using wrapper program. 

`func printVmDetails(namespace string) {
	kubeconfig := os.Getenv("KUBECONFIG")
	kv, _ := kubevirt.NewInstanceFromConfigFile(kubeconfig)
	listOptions := &k8smetav1.ListOptions{
		Limit: 1,
	}

	for {
		vlist, _ := kv.BatchListVirtualMachines(namespace, listOptions)

		for _, vm := range vlist.Items {
			fmt.Println(vm.Name, vm.Namespace, kv.IsVirtualMachineRunning(&vm))
			fmt.Println("Listing DVs:")
			printList(kv.GetVMDataVolumes(&vm))
			fmt.Println("Listing PVCs:")
			printList(kv.GetVMPersistentVolumeClaims(&vm))
			fmt.Println("Listing ConfigMaps:")
			printList(kv.GetVMConfigMaps(&vm))
			fmt.Println("Listing Secrets:")
			printList(kv.GetVMSecrets(&vm))
			if kv.IsVMAgentConnected(&vm) {
				if isPodRunning(vm, kv.GetVMPodLabel(&vm)) {
					fmt.Println("VM Pod is in running state displaying Freeze Thaw rules")
					fmt.Println(kv.GetVMFreezeRule(&vm))
					fmt.Println(kv.GetVMUnFreezeRule(&vm))
				}
			}

		}
		if vlist.Continue == "" {
			break
		} else {
			listOptions.Continue = vlist.Continue
			fmt.Println("Next Batch:")
		}
	}
}`

Screenshot 2023-11-29 at 2 42 22 PM

### DIFF
--- a/k8s/kubevirt/virtualmachine.go
+++ b/k8s/kubevirt/virtualmachine.go
@@ -9,6 +9,15 @@ import (
 	kubevirtv1 "kubevirt.io/api/core/v1"
 )
 
+const (
+	//VMPodSelectorLabel is the kubvirt label key for virt-launcher pod of the VM
+	VMPodSelectorLabel = "vm.kubevirt.io/name"
+	//VMFreezeCmd is the template for VM freeze command
+	VMFreezeCmd = "/usr/bin/virt-freezer --freeze --name %s --namespace %s"
+	//VMUnFreezeCmd is the template for VM unfreeze command
+	VMUnFreezeCmd = "/usr/bin/virt-freezer --unfreeze --name %s --namespace %s"
+)
+
 // VirtualMachineOps is an interface to perform kubevirt virtualMachine operations
 type VirtualMachineOps interface {
 	// CreateVirtualMachine calls VirtualMachine create client method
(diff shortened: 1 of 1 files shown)


k8s/kubevirt/virtualmachine.go
```diff
@@ -250,3 +275,35 @@ func (c *Client) GetVMConfigMaps(vm *kubevirtv1.VirtualMachine) []string {
...
+
+	return fmt.Sprintf(VMUnFreezeCmd, vm.GetName(), vm.GetNamespace())
+}
+
+// GetVMPodLabel return podSelector label for the given VM
+func (c *Client) GetVMPodLabel(vm *kubevirtv1.VirtualMachine) map[string]string {
+	return map[string]string{
+		VMPodSelectorLabel: vm.GetName(),
+	}
+}
```
mkoppal-px, line 309: Do we need these functions in sched-ops? I was thinking these are not using the receiver type `(c *Client)` and also these can be just a helper function in Torpedo itself right?
Shouldn't we be writing only those function here which use the kubevirt client?

vsundarraj-px: For phase 2, px-backup will consume this.
For regression cases with phase1 automation and for namespace backup scenario testing even for phase 2 test automation and integration test automation torpedo and px-backup will use it
Since this is tightly coupled with VM and signature should be same when called by any of them. Its better we use common function. Hence sched-ops seem to be the right place holder even though we are not making client call.

vsundarraj-px: For phase 2, px-backup will consume this.
For regression cases with phase1 automation and for namespace backup scenario testing even for phase 2 test automation and integration test automation torpedo and px-backup will use it
Since this is tightly coupled with VM and signature should be same when called by any of them. Its better we use common function. Hence sched-ops seem to be the right place holder even though we are not making client call.

prashanthpx: My view on this is for the following calls should they be in sched-ops or Torpedo? As these don't really need a client hanlder? We can discuss on this and see whats the best place to add this

GetVMFreezeRule()
 GetVMPodLabel()
 GetVMUnFreezeRule()

Following can be in sched-ops
IsVMAgentConnected()

vsundarraj-px:
I am closing this PR as commented these calls dont need client call and hence moved to stork where its heavily used. 
For more details see PR 
https://github.com/libopenstorage/stork/pull/1609

PR #355, opened 2024-02-13
PWX-36080 - Added files for policy to use apiversion v1

svijaykumar-px:
**What this PR does / why we need it**: Currently the poddisruptionbudget uses api version policy/v1beta1. But this was removed from kubernetes version 1.25 ( [refer to this link](https://kubernetes.io/docs/reference/using-api/deprecation-guide/#poddisruptionbudget-v125) ) . It was decided to keep the existing API as well.

Closes # PWX-36080

**Special notes for your reviewer**: This change is necessary to write operator integration tests for poddisruptionbudget as the old api version does not work.


k8s/policyV1/poddisruptionbudgetV1.go
piyush-nimbalkar, line 1: - Let's keep all the policy APIs in one folder under `k8s/policy`.
- Let's rename the old Ops and file as PodDisruptionBudgetV1Beta1Ops and poddisruptionbudget_v1beta1.go respectively. Also, let's rename the methods in it to have a V1Beta1 suffix just like the apiextensions package.
- Add a new file called poddisruptionbudget.go and have the v1 Ops and APIs in it without any suffix.
- People who vendor the latest sched-ops will rename their old methods to have the V1Beta1 suffix or change their API version to the newer one.

piyush-nimbalkar (changes requested):

k8s/policyV1/poddisruptionbudgetV1.go
svijaykumar-px: We cannot use the same policy.go file for both the versions as they use different client interfaces

k8s/policy/poddisruptionbudget_v1beta1.go
```diff
@@ -0,0 +1,67 @@
+package policy
+
+import (
+	"context"
+
+	policyv1beta1 "k8s.io/api/policy/v1beta1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// PodDisruptionBudgetOps is an interface to perform k8s Pod Disruption Budget operations
```
piyush-nimbalkar, line 10: nit: V1Beta1Ops

k8s/policy/policy.go
```diff
@@ -6,7 +6,7 @@ import (
 	"sync"
 
 	"github.com/portworx/sched-ops/k8s/common"
-	policyv1beta1client "k8s.io/client-go/kubernetes/typed/policy/v1beta1"
+	policyclient "k8s.io/client-go/kubernetes"
```
piyush-nimbalkar, line 9: nit: We don't have to call it policyclient. It is a Kubernetes client with many APIs. So let's leave it as kubernetes.

k8s/policy/policy.go
```diff
@@ -73,7 +74,7 @@ func NewInstanceFromConfigFile(config string) (Ops, error) {
 // Client is a wrapper for the kubernetes policy client.
 type Client struct {
 	config *rest.Config
-	policy policyv1beta1client.PolicyV1beta1Interface
+	policy policyclient.Interface
```
piyush-nimbalkar, line 77: nit: Let's rename the variable as just 'client' instead of policy now.

piyush-nimbalkar (approved):

PR #374, opened 2024-04-24
PWX-37013: WatchNode should watch all the nodes when no node is passed to the function

yjoshi-px:
What this PR does / why we need it:
Changed implementation of WatchNode function such that if no node is passed as an input it should watch all the nodes and should not return error

Which issue(s) this PR fixes (optional)
Closes # PWX-37013

Testing:
Screenshot 2024-04-24 at 9 57 52 AM


pnookala-px (approved):
Please add a comment that when we pass in nil, the watch watches all nodes.

PR #375, opened 2024-04-24
check VM status Ready along with Printable status

Rohit-PX:
* Current way of checking if VM is ready is incorrect
* VM status should also be checked to be `Ready` to certify that a VM is running

Closes #


dbhatnagar-px (approved):
Lgtm

PR #380, opened 2024-05-02
PWX-34271: handle a failed VMI migration with no migrationState

pureneelesh:

If the virt-launcher pod fails to get scheduled, VMI migration phase is changed to "Failed" but no migrationState is added to status. We were relying solely on migrationState to detect if the migration has completed. Now, we look at the Phase as well.

PWX-34271

### DIFF
--- a/k8s/kubevirt-dynamic/virtualmachineinstancemigration.go
+++ b/k8s/kubevirt-dynamic/virtualmachineinstancemigration.go
@@ -214,14 +214,23 @@ func (c *Client) unstructuredGetVMIMigration(
 		return nil, fmt.Errorf("failed to get 'phase' from the vmi migration status: %w", err)
 	}
 	// completed
-	ret.Completed, _, err = unstructured.NestedBool(migration.Object, "status", "migrationState", "completed")
+	var found bool
+	ret.Completed, found, err = unstructured.NestedBool(migration.Object, "status", "migrationState", "completed")
 	if err != nil {
 		return nil, fmt.Errorf("failed to get 'completed' from the vmi migration status: %w", err)
 	}
+	// migrationState is not present if the pod fails to get scheduled; we need to look at the Phase
+	if !found && ret.Phase == "Failed" {
+		ret.Completed = true
+	}
 	// failed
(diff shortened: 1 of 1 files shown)


k8s/kubevirt-dynamic/virtualmachineinstancemigration.go
```diff
@@ -214,15 +214,24 @@ func (c *Client) unstructuredGetVMIMigration(
...
+	}
 	// failed
-	ret.Failed, _, err = unstructured.NestedBool(migration.Object, "status", "migrationState", "failed")
+	ret.Failed, found, err = unstructured.NestedBool(migration.Object, "status", "migrationState", "failed")
 	if err != nil {
 		return nil, fmt.Errorf("failed to get 'failed' from the vmi migration status: %w", err)
 	}
+	// migrationState is not present if the pod fails to get scheduled; we need to look at the Phase
+	if !found && ret.Phase == "Failed" {
+		ret.Failed = true
```
pp511, line 233: If migration fails and no migrationState is added to status, both the above NestedBool calls will fail. In that case we will end up setting Completed=true but Failed=false(default) which will mean migration completed successfully which is not right.

pp511 (changes requested):

k8s/kubevirt-dynamic/virtualmachineinstancemigration.go
pp511: My bad. We are setting Failed to true from the Phase.

pp511 (approved):
Ran through all possible scenarios I could think of to find a flaw in this logic. Looks good.

k8s/kubevirt-dynamic/virtualmachineinstancemigration.go
pureneelesh: if migrationState is not present, err is nil and found if false. Additionally, if the phase is "Failed", we will set both ret.Completed and ret.Failed to true. I don't see why Failed will be default in that case.

pureneelesh: Ok, just saw your followup comment.

PR #382, opened 2024-05-09
PWX-36965: added method to watch events

pureneelesh:

PWX-36965

### DIFF
--- a/k8s/core/events.go
+++ b/k8s/core/events.go
@@ -3,6 +3,7 @@ package core
 import (
 	"context"
 
+	"github.com/sirupsen/logrus"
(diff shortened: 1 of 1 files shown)


Rohit-PX (approved):
To confirm, the options for calling this function, like the frequency etc will be left to the user right ?

pureneelesh:
> To confirm, the options for calling this function, like the frequency etc will be left to the user right ?

This function works similar to the existing watches on pods etc. You specify a callback function that is invoked every time there is a watch update. Standard ListOptions (labels etc) are honored by the watch.

PR #386, opened 2024-05-12
check if the VM is "startable" before trying to start it

pureneelesh:

This fixes intermittent failures with "always does not support manual start requests" error.

Closes #

### DIFF
--- a/k8s/kubevirt/virtualmachine.go
+++ b/k8s/kubevirt/virtualmachine.go
@@ -135,11 +135,17 @@ func (c *Client) ValidateVirtualMachineRunning(name, namespace string, timeout, 
 		return fmt.Errorf("failed to get Virtual Machine")
 	}
 
-	// Start the VirtualMachine if its not Started yet
-	if vm.Status.PrintableStatus == kubevirtv1.VirtualMachineStatusStopped ||
-		vm.Status.PrintableStatus == kubevirtv1.VirtualMachineStatusStopping {
+	// Start the VirtualMachine if it is "startable" and its not Started yet
+	// Avoid "Always does not support manual start requests" error by checking the RunStrategy
+	// https://bugzilla.redhat.com/show_bug.cgi?id=1709794
+	runStrategy, err := vm.RunStrategy()
+	if err != nil {
+		return fmt.Errorf("failed to get RunStrategy for VM %s/%s: %w", namespace, name, err)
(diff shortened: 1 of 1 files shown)


k8s/kubevirt/virtualmachine.go
```diff
@@ -135,25 +135,32 @@ func (c *Client) ValidateVirtualMachineRunning(name, namespace string, timeout,
...
-	if vm.Status.PrintableStatus == kubevirtv1.VirtualMachineStatusStopped ||
-		vm.Status.PrintableStatus == kubevirtv1.VirtualMachineStatusStopping {
+	// Start the VirtualMachine if it is "startable" and its not Started yet
+	// Avoid "Always does not support manual start requests" error by checking the RunStrategy
+	// https://bugzilla.redhat.com/show_bug.cgi?id=1709794
+	runStrategy, err := vm.RunStrategy()
+	if err != nil {
+		return fmt.Errorf("failed to get RunStrategy for VM %s/%s: %w", namespace, name, err)
+	}
+	if runStrategy != kubevirtv1.RunStrategyAlways && (vm.Status.PrintableStatus == kubevirtv1.VirtualMachineStatusStopped ||
```
vsundarraj-px, line 145: Validate call should actually just verify if the vm is running. Starting the vm should have been the independent call. 

This additional check with runStrategyAlways check will make this call invalid for this type. How will we check VM is running?

pureneelesh: Error on line 143 is returned only if the VM spec is invalid (check out the implementation of RunStrategy) which should not happen for us.

This change will fall through and do the check for whether the VM is ready regardless.

Only change is to skip `StartVirtualMachine()` when it is not really needed.

vsundarraj-px: Instead of here, we should skip return error in line 148 if startVM fails for Vm with runStrategyAlways. 
Although this check seems safe for me but not allowing starting of the VM itself doesnt seem to be right. And more so its intermittent so lets handle error for startVM

vsundarraj-px: Ok we checked online.. The runStrategyAlways is a right check to be in place if we want to start the VM. Change looks good to me

vsundarraj-px (approved):
LGTM

PR #390, opened 2024-05-29
PWX-37466: handle missing migrationState for a succeeded VM live-migr…

pureneelesh:
…ation

migrationState can be missing even for a successful live migration. If the phase is "Succeeded", we mark the migration as completed.

PWX-37466


pp511 (approved):
lgtm

PR #403, opened 2024-08-08
PWX-38340: fix race issues when updating k8s configmaps

pureneelesh:

Problems that this patch tries to fix:
1. Some callers updated the keys that they had not locked because they specified all keys to Patch()
2. Update() was dangerous because it could overwrite the internal keys used to store locks
3. RefreshLock() could lose the lock and re-acquire it siliently thereby overwriting the intervening update made by someone else
4. Patch could update the configmap while RefreshLock() has lost the ownership of the lock
5. V2 locks were re-entrant but did not keep the reference count causing premature unlocking

Changes:
1. Get rid of the configmap Update() method since it is too dangerous
2. Refactor Patch() into PatchKeyLocked() that updates a single key after verifying lock owner
3. RefreshLock detects loss of lock ownership
4. Add DeleteKeyLocked() method so that a key can be deleted safely
5. Add a generation number to the configmap and log each time the generation is incremented
6. Make v2 locks non-reentrant (like v1 locks) and fix one caller that was expecting the lock to be re-entrant

PWX-38340

The production code changes are being reviewed via another PR. Mainly review the unit tests here.

### DIFF
--- a/k8s/core/configmap/configmap.go
+++ b/k8s/core/configmap/configmap.go
@@ -1,7 +1,9 @@
 package configmap
 
 import (
+	"errors"
 	"fmt"
+	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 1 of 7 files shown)


adityadani (approved):
lgtm

PR #404, opened 2024-08-14
PWX-38340: don't wait for the old refreshLock goroutine to exit

pureneelesh:
Instead, the old refreshLock goroutine will detect that the lock has been re-acquired and will exit automatically.

PWX-38340

### DIFF
--- a/k8s/core/configmap/configmap_lock_v1.go
+++ b/k8s/core/configmap/configmap_lock_v1.go
@@ -16,9 +16,6 @@ func (c *configMap) Lock(id string) error {
 func (c *configMap) LockWithParams(id string, holdTimeout time.Duration, numAttempts uint) error {
 	fn := "LockWithParams"
 
-	// wait for any previous refreshLockV1 goroutine to exit
-	c.kLockV1.wg.Wait()
-
 	if numAttempts == 0 {
 		// This is the same no. of times (300) we try while acquiring a kvdb lock
 		numAttempts = c.lockAttempts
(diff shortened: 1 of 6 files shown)


k8s/core/configmap/configmap_lock_v1.go
```diff
@@ -177,6 +171,13 @@ func (c *configMap) refreshLockV1(id string) {
 		select {
 		case <-refresh.C:
 			c.kLockV1.Lock()
+			if c.kLockV1.v1LockGen != lockGen {
+				configMapLog(fn, c.name, id, "", nil).Infof(
+					"Lock generation has changed from %d to %d. Exiting from this refresh lock goroutine...",
+					lockGen, c.kLockV1.v1LockGen)
+				c.kLockV1.Unlock()
```
nrevanna, line 178: Would we want to painc here? 
Difference would mean re-entrant lock. Should our approach be soft and just log or panic and be assertive about it?

pureneelesh: This does not mean that the lock was reentered. We don't allow re-entrant lock. That is handled in the Lock() itself. This means that we are the previous refreshLock goroutine and we did not exit yet.

k8s/core/configmap/configmap_lock_v2_test.go
```diff
@@ -184,6 +186,58 @@ func TestLockHoldTimeout(t *testing.T) {
...
+	for i := 0; i < 100; i++ {
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			err = cm.LockWithKey(id1, key1)
+			require.NoError(t, err, "Unexpected error in LockWithKey(id1,key1)")
+
+			// give some time for the refreshLock goroutine to start
+			time.Sleep(time.Duration(rand.Intn(2000)) * time.Millisecond)
+
```
adityadani, line 230: How about you call PatchKeyLocked() here and check the contents of the config map, before you unlock ?

adityadani (approved):
lgtm

nrevanna (approved):

PR #408, opened 2024-09-13
PWX-38936 Don't return error if try to lock self-locked key

ssz1997:
Suppose a key is locked by a owner before it expires. The owner asks whether the key is locked, we check whether a goroutine that refreshes the lock expiration time exists. If yes, we tell it it's locked by you. If the goroutine doesn't exist, we tell it it's unlocked. So the "owner" will try to grab the lock in which a goroutine will be spawned. During this time, if someone else asks whether the key is locked, we turn them away saying it's locked by someone else.

https://purestorage.atlassian.net/browse/PWX-38936

### DIFF
--- a/k8s/core/configmap/configmap_lock_v2.go
+++ b/k8s/core/configmap/configmap_lock_v2.go
@@ -186,6 +186,8 @@ func (c *configMap) IsKeyLocked(key, requester string) (bool, string, error) {
 		c.kLocksV2Mutex.Lock()
 		defer c.kLocksV2Mutex.Unlock()
 		lock := c.kLocksV2[key]
+		lock.Lock()
+		defer lock.Unlock()
 		if requester == owner && (lock == nil || !lock.refreshing) {
 			return false, owner, nil
 		}
--- a/k8s/core/configmap/configmap_lock_v2.go
+++ b/k8s/core/configmap/configmap_lock_v2.go
@@ -66,6 +66,7 @@ func (c *configMap) LockWithKey(owner, key string) error {
 
 	configMapLog(fn, c.name, owner, key, nil).Debugf("Starting lock refresh")
 	go c.refreshLock(owner, key)
+	lock.refreshing = true
 	return nil
 }
 
@@ -154,7 +155,7 @@ func (c *configMap) UnlockWithKey(key string) error {
 	return err
 }
 
-func (c *configMap) IsKeyLocked(key string) (bool, string, error) {
+func (c *configMap) IsKeyLocked(key, requester string) (bool, string, error) {
 	// Get the existing ConfigMap
 	cm, err := core.Instance().GetConfigMap(
 		c.name,
@@ -174,13 +175,17 @@ func (c *configMap) IsKeyLocked(key string) (bool, string, error) {
 	}
 
 	if owner, ok := lockIDs[key]; ok {
-		// Someone owns the lock: let's check the expiration and the owner
+		// Existing key is unlocked if
+		//   1. Lock has expired; or
+		//   2. Lock owned by itself but goroutine refreshing expiration time doesn't exist
 		expiration := lockExpirations[key]
 		if time.Now().After(expiration) {
-			// Lock is expired
 			return false, "", nil
 		}
-
+		lock := c.kLocksV2[key]
+		if requester == owner && (lock == nil || !lock.refreshing) {
+			return false, owner, nil
+		}
 		return true, owner, nil
 	}
 
@@ -366,6 +371,7 @@ func (c *configMap) refreshLock(id, key string) {
 	}
 
 	startTime = time.Now()
+	defer func() { lock.refreshing = false }()
 	for {
 		select {
 		case <-refresh.C:
--- a/k8s/core/configmap/configmap_lock_v2.go
+++ b/k8s/core/configmap/configmap_lock_v2.go
@@ -184,8 +184,8 @@ func (c *configMap) IsKeyLocked(key, requester string) (bool, string, error) {
 			return false, "", nil
 		}
 		c.kLocksV2Mutex.Lock()
-		defer c.kLocksV2Mutex.Unlock()
 		lock := c.kLocksV2[key]
+		c.kLocksV2Mutex.Unlock()
 		lock.Lock()
 		defer lock.Unlock()
 		if requester == owner && (lock == nil || !lock.refreshing) {
--- a/k8s/core/configmap/configmap_lock_v2.go
+++ b/k8s/core/configmap/configmap_lock_v2.go
@@ -66,7 +66,6 @@ func (c *configMap) LockWithKey(owner, key string) error {
 
 	configMapLog(fn, c.name, owner, key, nil).Debugf("Starting lock refresh")
 	go c.refreshLock(owner, key)
-	lock.refreshing = true
 	return nil
 }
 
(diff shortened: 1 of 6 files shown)


k8s/core/configmap/configmap_lock_v2.go
```diff
@@ -66,6 +66,7 @@ func (c *configMap) LockWithKey(owner, key string) error {
 
 	configMapLog(fn, c.name, owner, key, nil).Debugf("Starting lock refresh")
 	go c.refreshLock(owner, key)
+	lock.refreshing = true
```
nrevanna, line 69: This is an indication of the routine's existence. Inside of refreshLock would be the right place for this assignment.

k8s/core/configmap/configmap_lock_v2.go
```diff
@@ -174,13 +175,17 @@ func (c *configMap) IsKeyLocked(key string) (bool, string, error) {
...
+		// Existing key is unlocked if
+		//   1. Lock has expired; or
+		//   2. Lock owned by itself but goroutine refreshing expiration time doesn't exist
 		expiration := lockExpirations[key]
 		if time.Now().After(expiration) {
-			// Lock is expired
 			return false, "", nil
 		}
-
+		lock := c.kLocksV2[key]
```
nrevanna, line 185: don't you need to hold kLocksV2Mutex lock for this map?

k8s/core/configmap/configmap_lock_v2.go
```diff
@@ -174,13 +175,17 @@ func (c *configMap) IsKeyLocked(key string) (bool, string, error) {
 	}
 
 	if owner, ok := lockIDs[key]; ok {
-		// Someone owns the lock: let's check the expiration and the owner
+		// Existing key is unlocked if
+		//   1. Lock has expired; or
+		//   2. Lock owned by itself but goroutine refreshing expiration time doesn't exist
 		expiration := lockExpirations[key]
 		if time.Now().After(expiration) {
-			// Lock is expired
```
nrevanna, line 180: Why remove the comment?

k8s/core/configmap/configmap_lock_v2.go
```diff
@@ -174,13 +175,17 @@ func (c *configMap) IsKeyLocked(key string) (bool, string, error) {
 	}
 
 	if owner, ok := lockIDs[key]; ok {
-		// Someone owns the lock: let's check the expiration and the owner
+		// Existing key is unlocked if
+		//   1. Lock has expired; or
+		//   2. Lock owned by itself but goroutine refreshing expiration time doesn't exist
```
nrevanna, line 180: nit: goroutine refreshing expiration time doesn't exist -> refresh goroutine is not running

It might be worth pointing out somewhere in the comment that IsKeyLocked can return locked for the non-owners and not-locked for the owner. May be comment on top of the function defition.

k8s/core/configmap/configmap_lock_v2_test.go
```diff
@@ -402,3 +402,72 @@ func TestCMLockTTL(t *testing.T) {
...
+	key := "is-key-locked-key"
+
+	locked, _, err := cm.IsKeyLocked(key, id1)
+	require.NoError(t, err)
+	require.False(t, locked)
+
+	// lock the key
+	err = cm.LockWithKey(id1, key)
+	require.NoError(t, err)
+	require.True(t, cm.kLocksV2[key].refreshing)
```
nrevanna, line 423: Map access outside of lock might throw concurrent map access error. Which could make this test flaky.

k8s/core/configmap/configmap_lock_v2_test.go
```diff
@@ -402,3 +402,72 @@ func TestCMLockTTL(t *testing.T) {
...
+	require.NoError(t, err)
+	require.True(t, locked)
+	require.Equal(t, id1, owner)
+	locked, owner, err = cm.IsKeyLocked(key, id2)
+	require.NoError(t, err)
+	require.True(t, locked)
+	require.Equal(t, id1, owner)
+
+	// stop goroutine and make lock expired
+	close(cm.kLocksV2[key].done)
```
nrevanna, line 436: lock needed

k8s/core/configmap/configmap_lock_v2_test.go
```diff
@@ -402,3 +402,72 @@ func TestCMLockTTL(t *testing.T) {
...
+	cm.kLocksV2[key].unlocked = true
+	time.Sleep(2 * testLockTTL)
```
nrevanna, line 438: You don't need to sleep here. You can patch time and fast forward to what time you need.
you can keep a variable patchedTime and make the patched function return it. Before the next step, you can assign to patchedTime what value you expect. And it can be done multiple times in the same test.

sangleganesh:
imo this looks like a lot of code and complexity for some micro optimization. it will also encourage users to not define locks to protect resources properly and end up having double locks work until we get a surprise.

k8s/core/configmap/configmap_lock_v2.go
ssz1997: The explanation is moved a few lines above

k8s/core/configmap/configmap_lock_v2.go
ssz1997: Makes sense. We are not be able to check this value gets updated in the test anymore because the check may happen before the refreshing value gets assigned.

adityadani:
I don't think this is a micro optimization. The lock functionality and the way it is used is currently broken and thats what this PR tries to fix. 

IMHO if we want to avoid the callers using this lock functionality incorrectly and encourage them to protect resources properly, we should get rid of the IsKeyLocked() function and solely rely on LockWithKey(). LockWithKey should do the existing checks of owner and expiration but at the same time also check the existence of refresh routine. 

I am fine with this being done as a follow up PR to unblock 3.2.0

adityadani:
@ssz1997 What I meant is lets merge this PR, and we can get rid of the IsKeyLocked() function later on to keep the lock implementation clear and simple.

k8s/core/configmap/configmap_lock_v2.go
```diff
@@ -174,13 +176,19 @@ func (c *configMap) IsKeyLocked(key string) (bool, string, error) {
...
+		//   1. Lock has expired; or
+		//   2. Lock owned by itself but refresh goroutine is not running
 		expiration := lockExpirations[key]
 		if time.Now().After(expiration) {
-			// Lock is expired
 			return false, "", nil
 		}
-
+		c.kLocksV2Mutex.Lock()
+		defer c.kLocksV2Mutex.Unlock()
```
nrevanna, line 187: - It's not necessary defer the unlock which will hold the lock until the end of the funciton. You can unlock after line 188
- In the refreshLock function, you are using lock.Lock() to update refreshing. It'll be more appropriate to use the samelock for access in line 189. That unlock you could defer

nrevanna (commented):
LGTM after addressing the one comment I have.

k8s/core/configmap/configmap_lock_v2.go
ssz1997: I'm not understanding the second point. How to use the same lock in line 189 in IsKeyLocked() in refreshLock()?

nrevanna: Clarified offline

nrevanna (approved):

ssz1997:
```
--- FAIL: TestDeleteConfigMap (0.29s)
    configmap_test.go:103: Deleting configmap: kube-system/px-configmaps-test
    configmap_test.go:42: 
        	Error Trace:	/home/travis/build/portworx/sched-ops/k8s/core/configmap/configmap_test.go:42
        	Error:      	Received unexpected error:
        	            	configmaps "px-configmaps-test" not found
        	Test:       	TestDeleteConfigMap
        	Messages:   	Unexpected error in delete
```

jrivera-px (approved):
LGTM, thanks...

PR #413, opened 2024-10-14
[PWX-38627] Stop logging Conflict error for configmap lock

ssz1997:
Since we already have warning log if it takes 1.5 times of the refresh duration, the conflict error logging can be too spammy. Thus stop logging this error.

Closes #38627

### DIFF
--- a/k8s/core/configmap/configmap_lock_v2.go
+++ b/k8s/core/configmap/configmap_lock_v2.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"math/rand"
 	"time"
 
 	v1 "k8s.io/api/core/v1"
@@ -400,7 +401,8 @@ func (c *configMap) refreshLock(id, key string) {
 									" [ID %v] [Key %v] [Err: %v] [Current Refresh: %v] [Previous Refresh: %v]",
 								isConflictErrCount, id, key, err, currentRefresh, prevRefresh)
 						}
-						// try refreshing again
+						// try refreshing again after a random sleep so nodes wouldn'e entangle for too long
+						time.Sleep(lockSleepDuration + time.Duration(rand.Intn(lockRandomSleepDurationMaxMillisecond))*time.Millisecond)
 						continue
 					}
 					configMapLog(fn, c.name, "", key, err).Errorf(
--- a/k8s/core/configmap/configmap_lock_v2.go
+++ b/k8s/core/configmap/configmap_lock_v2.go
@@ -392,15 +392,15 @@ func (c *configMap) refreshLock(id, key string) {
 				c.checkLockTimeout(c.defaultLockHoldTimeout, startTime, id)
 				currentRefresh = time.Now()
 				if _, err := c.tryLock(id, key, true); err != nil {
+					if k8s_errors.IsConflict(err) {
+						// try refreshing again
+						continue
+					}
 					configMapLog(fn, c.name, "", key, err).Errorf(
 						"Error refreshing lock. [ID %v] [Key %v] [Err: %v]"+
 							" [Current Refresh: %v] [Previous Refresh: %v]",
 						id, key, err, currentRefresh, prevRefresh,
 					)
-					if k8s_errors.IsConflict(err) {
-						// try refreshing again
-						continue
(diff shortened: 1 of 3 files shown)


k8s/core/configmap/configmap_lock_v2.go
```diff
@@ -392,15 +392,15 @@ func (c *configMap) refreshLock(id, key string) {
 				c.checkLockTimeout(c.defaultLockHoldTimeout, startTime, id)
 				currentRefresh = time.Now()
 				if _, err := c.tryLock(id, key, true); err != nil {
+					if k8s_errors.IsConflict(err) {
+						// try refreshing again
+						continue
+					}
```
pp511, line 398: We might encounter multiple conflict errors in succession, with a success occurring just before lockRefreshDuration * 1.5. In such cases, it would be unclear if any conflict errors occurred. To balance between excessive logging and capturing if there are too many conflict errors, we should log the error if it happens more than a set number of times. Logging after 5 consecutive conflict errors seems like a reasonable threshold.

pp511: Sounds ok. And let's file a followup PWX to improve the logic(add backoff & or jitter in retrying) for the next release.

k8s/core/configmap/configmap_lock_v2.go
```diff
@@ -387,20 +387,27 @@ func (c *configMap) refreshLock(id, key string) {
...
 			lock.Lock()
-
+			isConflictErrCount := 0
 			for !lock.unlocked {
 				c.checkLockTimeout(c.defaultLockHoldTimeout, startTime, id)
 				currentRefresh = time.Now()
 				if _, err := c.tryLock(id, key, true); err != nil {
+					if k8s_errors.IsConflict(err) {
+						isConflictErrCount++
+						if isConflictErrCount%10 == 0 {
```
pp511, line 397: nit..Instead of hardcoding 10 let's use a global variable..numRetryCount or numConflictErrorRetryCount o

k8s/core/configmap/configmap_lock_v2.go
```diff
@@ -387,20 +387,27 @@ func (c *configMap) refreshLock(id, key string) {
...
+							configMapLog(fn, c.name, "", key, err).Errorf(
+								"Error refreshing lock due to conflict from concurrent configmap updates. retries: %v."+
+									" [ID %v] [Key %v] [Err: %v] [Current Refresh: %v] [Previous Refresh: %v]",
+								isConflictErrCount, id, key, err, currentRefresh, prevRefresh)
+						}
+						// try refreshing again
+						continue
```
pp511, line 404: I am thinking..it is possible that multiple nodes are stuck in this same state then they will continuously get conflict errors and we continuing from here will not help either. That is the current behavior too but we can improve this a little by adding a sleep (1sec + random jitter).

ssz1997: Makes a lot of sense

k8s/core/configmap/configmap_lock_v2.go
ssz1997: Very specific to a logging pattern so I'm keeping it hardcoded

k8s/core/configmap/configmap_lock_v2.go
```diff
@@ -387,20 +388,28 @@ func (c *configMap) refreshLock(id, key string) {
...
 				if _, err := c.tryLock(id, key, true); err != nil {
+					if k8s_errors.IsConflict(err) {
+						isConflictErrCount++
+						if isConflictErrCount%10 == 0 {
+							configMapLog(fn, c.name, "", key, err).Errorf(
+								"Error refreshing lock due to conflict from concurrent configmap updates. retries: %v."+
+									" [ID %v] [Key %v] [Err: %v] [Current Refresh: %v] [Previous Refresh: %v]",
+								isConflictErrCount, id, key, err, currentRefresh, prevRefresh)
+						}
+						// try refreshing again after a random sleep so nodes wouldn'e entangle for too long
```
pp511, line 404: nit..typo wouldn't

pp511 (approved):
just one very minor nitpick.
lgtm

PR #419, opened 2025-12-10
update k8s,go version and other dependencies

kg-px:
Updated K8s, go version to fix vul in Autopilot.

Closes # https://purestorage.atlassian.net/browse/PWX-49152

Ques to Repo owner: It would be great if we can have a release/tagged version of this PR, even if we don't want to take it to master.


kg-px:
this has been moved to pure-px repo

